PR #147, opened 2020-11-05
Story/143

MicheleK17:


cypress/integration/test_language.js
```diff
@@ -0,0 +1,19 @@
+describe("Test langugae setting in local storage", function () {
+    it("check english language storage", function () {
+        cy.visit("http://localhost:8080/");
```
dptolle, line 3: This is not going to work in a CI (i.e. Jenkins) setting

dptolle (changes requested):

cypress/integration/test_language.js
MicheleK17: Oh, yes I understand.

dptolle (approved):

PR #164, opened 2020-12-04
Add accordion with number of isolates

MicheleK17:
- count isolates (total and selected)

Ticket: #160
Change-Id: I9b5564046ec9a4e4909602f64e92a8be583aa8b9

### DIFF
--- a/src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
+++ b/src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
@@ -7,7 +7,6 @@ import AccordionSummary from "@material-ui/core/AccordionSummary";
 import AccordionDetails from "@material-ui/core/AccordionDetails";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import { DataContext } from "../../../../Shared/Context/DataContext";
-import { countNrOfIsolates } from "./countNrOfIsolates.component";
 
 const subHeadingStyle = css`
     margin: 0;
@@ -30,8 +29,8 @@ export function QueryPageNrOfIsolatesComponent(): JSX.Element {
     const { t } = useTranslation(["QueryPage"]);
     const { data } = useContext(DataContext);
 
-    const totalNrOfIsolates = countNrOfIsolates(data.ZNData);
-    const nrOfSelectedIsolates = countNrOfIsolates(data.ZNDataFiltered);
+    const totalNrOfIsolates = data.ZNData.length;
+    const nrOfSelectedIsolates = data.ZNDataFiltered.length;
 
     return (
         <Accordion defaultExpanded>
--- a/src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
+++ b/src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
@@ -1,0 +1,67 @@
+/** @jsx jsx */
+import { css, jsx } from "@emotion/core";
+import { useTranslation } from "react-i18next";
+import { useContext } from "react";
+import Accordion from "@material-ui/core/Accordion";
+import AccordionSummary from "@material-ui/core/AccordionSummary";
+import AccordionDetails from "@material-ui/core/AccordionDetails";
+import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
+import { DataContext } from "../../../../Shared/Context/DataContext";
+import { countNrOfIsolates } from "./countNrOfIsolates.component";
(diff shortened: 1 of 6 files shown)


src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -32,37 +32,35 @@ export function QueryPageParameterContentComponent(): JSX.Element {
...
-                                    );
-                                });
-                                return elements;
-                            })()}
-                        </tbody>
-                    </table>
-                </AccordionDetails>
-            </Accordion>
-        </div>
+        <Accordion defaultExpanded>
```
MicheleK17, line 35: Only the <div> was removed - the rest is not new code

src/app/Main/Body/QueryPage/NumberOfIsolates/countNrOfIsolates.component.ts
```diff
@@ -0,0 +1,5 @@
+import { DBentry } from "../../../../Shared/Isolat.model";
+
+export function countNrOfIsolates(data: DBentry[]): string {
+    return (Object.keys(data).length as unknown) as string;
```
GerberI, line 4: - why not "return String(data.length)"
- file is named like it is a component but it exports only a helper function
- i would remove this file because only purpose is to get the length of an array

src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+`;
+const tableNumberStyle = css`
+    padding-right: 5em;
+`;
+
+export function QueryPageNrOfIsolatesComponent(): JSX.Element {
+    const { t } = useTranslation(["QueryPage"]);
+    const { data } = useContext(DataContext);
+
+    const totalNrOfIsolates = countNrOfIsolates(data.ZNData);
```
GerberI, line 33: why not: "= data.ZNData.length" ?

src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+                aria-controls="panel1a-content"
+                id="panel1a-header"
+            >
+                <h3 css={subHeadingStyle}>{t("NrOfIsolates.Title")}</h3>
+            </AccordionSummary>
+            <AccordionDetails css={parameterBlockStyle}>
+                <table css={tableStyle}>
+                    <tbody>
+                        <tr>
+                            <td key="tableText-total" css={tableTextStyle}>
```
GerberI, line 49: whats that key attribute?

GerberI (changes requested):

src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
MicheleK17: Each table element needs a unique key, otherwise there will be a browser warning

src/app/Main/Body/QueryPage/NumberOfIsolates/countNrOfIsolates.component.ts
MicheleK17: Thanks, you are absolutely right.

GerberI (approved):

src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
GerberI: I can not find any documentation about the key attribute. Is this a feature of some library you use? Whats the exact message of the warning?

MicheleK17: The warning was "Warning: Each child in a list should have a unique "key" prop". 
But it seems that you are right and I need the key only for lists (material-ui) and not for tables. I don't remember why I used it here.

GerberI: Ok this is a react attribute:
https://reactjs.org/docs/lists-and-keys.html
There is a similar feature in our angular apps. It is for helping change detection to identifying unchanged elements. E.g. if you reorder a list react do not need to render the individual items because they are unchanged but only change the layout. I would consider reading into this further, there seems to be some misunderstanding how and when to use this feature and false usage can lead to bugs during rendering.

PR #172, opened 2020-12-15
Change table style to dark blue

MicheleK17:
Ticket: #159
Change-Id: I23e3a7a87d9fc05b7e1a7933e2d2e9a416a3ff54

### DIFF
--- a/src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
+++ b/src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
@@ -1,7 +1,8 @@
 /** @jsx jsx */
 import { css, jsx, SerializedStyles } from "@emotion/core";
 import {
-    backgroundColor,
+    bfrPrimaryPalette,
+    onPrimaryColor,
     primaryColor,
 } from "../../../../Shared/Style/Style-MainTheme.component";
 
@@ -12,13 +13,8 @@ const titleDivStyle = (
     display: ${isTitle ? "flex" : "none"};
     align-items: ${isRow ? "flex-end" : "center"};
     justify-content: ${isRow ? "center" : "flex-end"};
-    background-color: ${backgroundColor};
-    border-right: ${isRow
-        ? `double ${primaryColor}`
-        : `1px solid ${primaryColor}`};
-    border-bottom: ${isRow
-        ? `1px solid ${primaryColor}`
-        : `double ${primaryColor}`};
+    background-color: ${primaryColor};
+    border-bottom: ${isRow ? "none" : `solid ${bfrPrimaryPalette[300]}`}
 `;
 const tableTitleStyle = (
     isTitle: boolean,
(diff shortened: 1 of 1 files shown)


src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
```diff
@@ -29,13 +25,17 @@ const tableTitleStyle = (
     display: ${isTitle ? "flex" : "none"};
     width: ${isRow ? "inherit" : `${width}px`};
     height: ${isRow ? `${height}px` : "inherit"};
-    margin: 0;
-    padding: ${isRow ? "0 0.3em" : "0.2em 0"};
+    min-height: max-content;
```
GerberI, line 28: Consider evading working draft css attributes. They can cause compatibility issues. See:
https://developer.mozilla.org/en-US/docs/Web/CSS/min-height
In general check javascript/html/css features for compatibility before using them. MDN has a nice compatibility table for each feature.

src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
```diff
@@ -29,13 +25,17 @@ const tableTitleStyle = (
...
+    margin: ${isRow ? "0.3em 0" : "0"};
+    padding: ${isRow ? "0 0.2em" : "0.2em 0"};
     justify-content: center;
     align-items: center;
     font-weight: normal;
-    background-color: ${backgroundColor};
-    color: ${primaryColor};
+    background-color: ${primaryColor};
+    color: ${onPrimaryColor};
+    writing-mode: ${isRow ? "vertical-lr" : "none"};
```
GerberI, line 37: i think this attribute is meant for languages with different writing alignment:
https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode
Also you already rotate the element by 180 degree so this seems obsolete.

GerberI (changes requested):

src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
MicheleK17: I don't know how else to create the look I want. 
If I rotate the font only 90 degrees, then the font overlaps the div around it.

src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
MicheleK17: According to the table, "min-height: max-content" only causes problems when using internet explorer, so I see no reason not to use it?

GerberI: You right, but it is still a working draft. The behavior can change anytime and browsers can introduce bugs because of changes under the hood. I've had problems with the box sizing keywords myself and removed them. But you can use them if you are aware of the problems that might occur.

src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
GerberI: That's because transformations are done outside the layout rendering phase. I don't know how to do it correctly but the writing-mode + transform looks like a hack. If there is no better solution you can leave it as it is.

GerberI (approved):

src/app/Main/Body/QueryPage/Results/Result-MainHeader.component.tsx
MicheleK17: In other cases, absolute positioning is used, but this makes no sense in my case. I will talk to Katja and Dominic in the new year if we really need to rotate the font.

PR #176, opened 2020-12-16
Display Parameter in two columns

MicheleK17:
- remove table

Ticket: #169
Change-Id: I60b37de8e15880b580f9aa5062fd8c94a83a1d96

### DIFF
--- a/src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
+++ b/src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
@@ -8,11 +8,7 @@ import AccordionDetails from "@material-ui/core/AccordionDetails";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import _ from "lodash";
 import { FilterContext } from "../../../../Shared/Context/FilterContext";
-import {
-    FilterInterface,
-    FilterType,
-    mainFilterAttributes,
-} from "../../../../Shared/Filter.model";
+import { mainFilterAttributes } from "../../../../Shared/Filter.model";
 import { ParameterListComponent } from "./Parameter-List.component";
 
 const subHeadingStyle = css`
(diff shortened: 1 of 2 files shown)


GerberI (changes requested):
Some general consideration:
- Some function names are misleading. You use addSomething when you actually create react components, so use names like createSomeComponent in those cases
- Even better put those components in separate react component files
- You've splittet your Parameter layout into to components, one creates pairs and the other puts them to a flex box. It would be better to encapsulate this in only one component. E.g. your ParameterList could be responsible for putting all the parameters in the right layout.
- Why you even split the parameters in pairs? Why not put them all in a flex item and let the browser decide how many items to place in a row?

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -45,12 +45,12 @@ export function QueryPageParameterContentComponent(): JSX.Element {
         }
     });
 
-    const AddParameterElement = ((filterAttributes: FilterType[], filterValues: FilterInterface): JSX.Element[] => {
+    const createParameterComponent = (
```
GerberI, line 48: actually you create component[s]

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -45,12 +45,12 @@ export function QueryPageParameterContentComponent(): JSX.Element {
...
+        filterAttributes: FilterType[],
+        filterValues: FilterInterface
+    ): JSX.Element[] => {
         const elements: JSX.Element[] = [];
-        pairwise(filterAttributes, function func(
-            current: FilterType,
-            next: FilterType
-        ): void {
+        pairwise(filterAttributes, (current, next): void => {
```
GerberI, line 53: You can remove the "void" too.
To explain: If you declare a function you type it, if you use a function or define it like here, typescript will infer the types for you. If you are interested you can google function declaration vs function definition.
Another thought to it: Would you use something like: foo('xyz': string) instead of foo('xyz') ? foo already says it gets a string, so why typing it twice? It's the same like foo((x:string):void=>{}) vs foo((x)=>{})
Actually type inference is more complicated than i stated it here but it is only interesting for the guys who write the compilers (in most cases).

GerberI (approved):

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
MicheleK17: I will make a separate commit in the refactoring only for renaming variables. :-D

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
MicheleK17: I went with your idea of letting the browser decide how many elements are displaied in a row. 
Nevertheless, thanks for the input. I'll try to keep it in mind next time.

GerberI (approved):
Thank's for the change, looks better now

PR #179, opened 2020-12-17
Explanation Text if no row/column selected

MicheleK17:
- move Table from "ResultData" to "Content" component
    - remove check isIsolates from table styling

Ticket: #167
Change-Id: Id3d520d8a135e04588f8b22ab693b67037e50187

### DIFF
--- a/src/app/Main/Body/QueryPage/Results/QueryPage-ResultData.component.tsx
+++ b/src/app/Main/Body/QueryPage/Results/QueryPage-ResultData.component.tsx
@@ -9,7 +9,7 @@ import {
 } from "@material-ui/core";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import { TableContext } from "../../../../Shared/Context/TableContext";
-import { ResultsTableComponent as ResultsTable } from "./Results-Table.component";
+import { TableContentComponent as TableContent } from "./Result-Content.component";
 import { TableMainHeaderComponent } from "./Result-MainHeader.component";
 
 const subHeadingStyle = css`
@@ -22,9 +22,6 @@ const dataStyle = css`
     max-width: fit-content;
     margin: auto;
     box-sizing: inherit;
-`;
-const dataTableStyle = css`
-    overflow: auto;
 `;
 const tableDivStyle = css`
     display: flex;
@@ -88,19 +85,9 @@ export function QueryPageTableRestultComponent(
 
     const headerWidth: number = totalWidth - partWidth;
 
-    const accordionHeader: string = t(`Results.Table`)
+    const accordionHeader: string = t(`Results.Table`);
     const rowMainHeader: string = t(`Filters.${table.row}`);
     const colMainHeader: string = t(`Filters.${table.column}`);
-
-    const isIsolates = !!(
-        !props.displayRowCol.isCol && !props.displayRowCol.isRow
-    );
-    const isRowAndCol = !!(
-        props.displayRowCol.isCol && props.displayRowCol.isRow
-    );
-    const isRowNotCol = !!(
-        !props.displayRowCol.isCol && props.displayRowCol.isRow
-    );
 
     return (
         <Accordion defaultExpanded>
@@ -128,21 +115,11 @@ export function QueryPageTableRestultComponent(
                             width={headerWidth}
                             text={rowMainHeader}
                         />
-                        <div
-                            css={dataTableStyle}
-                            ref={(node: HTMLElement | null) =>
-                                div(node, "totalWidth")
-                            }
-                        >
-                            <ResultsTable
-                                allIsolates={table.statisticData}
-                                isIsolates={isIsolates}
-                                columnAttributes={props.columnAttributes}
-                                getSize={div}
-                                isRowNotCol={isRowNotCol}
-                                isRowAndCol={isRowAndCol}
-                            />
-                        </div>
+                        <TableContent
+                            displayRowCol={props.displayRowCol}
+                            columnAttributes={props.columnAttributes}
+                            getSize={div}
+                        />
                     </div>
                 </div>
             </AccordionDetails>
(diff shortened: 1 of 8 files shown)


src/app/Main/Body/QueryPage/Results/Result-Content.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+    overflow: auto;
+`;
+
+interface ContentInterface {
+    displayRowCol: {
+        isCol: boolean;
+        isRow: boolean;
+    };
+    columnAttributes: string[];
+    getSize: (
```
GerberI, line 18: reject: this does not get the size, it sets the state of a parent component. It bypasses the normal dataflow so at least the name should reflect that.
Consider removing this callback and try to solve the problem in more common way

src/app/Main/Body/QueryPage/Results/Result-Content.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+        node: HTMLElement | null,
+        key: "height" | "totalWidth" | "partWidth"
+    ) => void;
+}
+
+export function TableContentComponent(props: ContentInterface): JSX.Element {
+    const { table } = useContext(TableContext);
+
+    const noRowAndCol = !!(
```
GerberI, line 27: reject: remove !!, all you compare here are booleans, so no boolean conversion hack necessary.
Try to avoid such statements

src/app/Main/Body/QueryPage/Results/Result-Content.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+        !props.displayRowCol.isCol && props.displayRowCol.isRow
+    );
+
+    if (noRowAndCol) {
+        return <ExplanationText />;
+    }
+    return (
+        <div
+            css={dataTableStyle}
+            ref={(node: HTMLElement | null) =>
```
GerberI, line 43: Avoid direct DOM interactions:
https://reactjs.org/docs/refs-and-the-dom.html

src/app/Main/Body/QueryPage/Results/Results-Table.component.tsx
```diff
@@ -82,7 +79,6 @@ export function ResultsTableComponent(props: ResultTableProps): JSX.Element {
                             {RowValues(
                                 row,
                                 classes,
-                                props.isIsolates,
                                 props.isRowAndCol,
                                 tabelCellStyle
```
GerberI, line 83: Avoid using callbacks as props. React is up -> down not down -> up the hierarchy. By using callbacks you could easily break all kinds of stuff and introduce side effects to your code which could become uncontrollable in a later state of the app.

GerberI (changes requested):
You should establish a more strict naming convention. It is hard to understand the code because you name things in different ways. Naming things is the foundation of good code. E.g.
- FooProps vs fooInterface for react props
- getFoo when you actually meant setFoo
- foo when you meant a function getFoo
- noRowAndCol vs isRowAndCol vs isRowNotCol
- Foo-Bar vs FooBar
- Foo vs FooComponent for react components
- common names instead of explicit names (e.g. columnAttributes or ContentInterface)
- mismatching file and component names

You use a very jerky pattern in these components. You tell your parents to recalculate their size from some child components using a callback (getSize). Please consider removing this pattern. You bypass and interfer with the react rendering and browser rendering phases. This can lead to a lot of problems. You should rely on the browser to render things in their correct sizes by using only native css styling like flex. These manual size calculations by interacting directly with the dom was the reason why we removed the handson table from the mibi-portal. It used a similar mechanism to calculate its size and messed up the whole renderingleading to a lot of undesired side effects and hard to trace bugs.

GerberI (approved):

src/app/Main/Body/QueryPage/Results/QueryPage-ResultData.component.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from "@material-ui/core";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import { TableContext } from "../../../../Shared/Context/TableContext";
```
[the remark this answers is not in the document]
MicheleK17: table of the TableContext is used for the useCallback function

src/app/Main/Body/QueryPage/Results/Result-Content.component.tsx
MicheleK17: I already tried to solve it in another way. But I didn't get the result that the font is centered over the corresponding rows/columns (and not the whole space) AND only the inner table is scrolled in case of an overflow.

src/app/Main/Body/QueryPage/Results/QueryPage-ResultData.component.tsx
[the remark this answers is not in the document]
GerberI: ack

src/app/Main/Body/QueryPage/Results/Result-Content.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+        isRow: boolean;
+    };
+    columnAttributes: string[];
+    getSize: (
+        node: HTMLElement | null,
+        key: "height" | "totalWidth" | "partWidth"
+    ) => void;
+}
+
+export function TableContentComponent(props: ContentInterface): JSX.Element {
```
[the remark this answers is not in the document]
GerberI: done

GerberI (approved):

src/app/Main/Body/QueryPage/Results/Result-Content.component.tsx
GerberI: if you like we can discuss this issue next year. Or i can try and fix this for you. But it would be could to create a refactoring ticket for this.

MicheleK17: Yes discussing it next year would be great! I make a ticket.

GerberI: thx

PR #188, opened 2021-01-18
Rename remove tooltip

MicheleK17:
- "alles abwählen" zu "Auswahl zurücksetzen"

Ticket: #187
Change-Id: Ib011eaf6a6e974e9ea5c74582bef7b75bad8b546


src/locales/de/QueryPage.json
```diff
@@ -11,7 +11,7 @@
     },
     "Buttons": {
         "Submit": "Ergbenisse anzeigen",
```
GerberI, line 13: here is a typo

GerberI (changes requested):

src/locales/de/QueryPage.json
MicheleK17: I removed the entire entry because it is no longer needed.

GerberI (approved):

PR #194, opened 2021-01-19
Implement table option bar 

MicheleK17:


src/app/Core/calculateRowsWithPercent.service.ts
```diff
@@ -0,0 +1,43 @@
+function getSumOfIsolates(objList: Record<string, string>[]): number {
+    let sum = 0;
+    objList.forEach((obj) => {
+        const k = Object.keys(obj);
+        k.forEach((element) => {
+            if (element !== "name") {
+                sum += (obj[element] as unknown) as number;
```
GerberI, line 7: reject: use something like:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
telling typescript to ignore the type doesn't prevent your app from crashing if something goes wrong.

src/app/Core/calculateRowsWithPercent.service.ts
```diff
@@ -0,0 +1,43 @@
...
+            if (element !== "name") {
+                sum += (obj[element] as unknown) as number;
+            }
+        });
+    });
+    return sum;
+}
+
+function calculatePercent(part: number, total: number): string {
+    const percent = (part / total) * 100;
```
GerberI, line 15: reject: handle the case if total = 0, to prevent this function from throwing an exception. Or if 0 is forbidden explicitly throw an exception if total = 0

src/app/Core/calculateRowsWithPercent.service.ts
```diff
@@ -0,0 +1,43 @@
...
+): Record<string, string>[] {
+    const rowsWithPercent: Record<string, string>[] = [];
+    const sumOfIsolates: number = getSumOfIsolates(objList);
+    objList.forEach((obj) => {
+        const percentRow: Record<string, string> = { name: obj.name };
+        const k = Object.keys(obj);
+        k.forEach((element) => {
+            if (element !== "name") {
+                percentRow[element] = calculatePercent(
+                    (obj[element] as unknown) as number,
```
GerberI, line 35: same as above

GerberI (changes requested):
Commit: Change TableContext and calculate percentage
consideration: i would say this is not a core service, it should be where the other stuff regarding the table data is.
consideration: try to evade names such as objList, element, obj. The service is for doing stuff with your table result data, so i would name things accordingly.
consideration: try to use map, reduce or lodash for manipulating arrays and objects. So e.g. a foreach(obj) objects.push(obj) is actually a map function. Or summing up something could be done with a reduce function

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -103,7 +110,7 @@ export function ContentResultsComponent(): JSX.Element {
             setColumnAttributes([t("Results.TableHead")]);
             setTable({
                 ...table,
-                statisticData: [
+                statisticDataAbsolute: [
```
GerberI, line 113: reject: you should set statisticDataPercentage to a default too?

src/app/Main/Body/QueryPage/Results/TableResults/TableResults-Options.component.tsx
```diff
@@ -0,0 +1,61 @@
...
+`;
+const BlueRadio = withStyles(() =>
+    createStyles({
+        root: {
+            color: bfrDarkgrey,
+            "&$checked": {
+                color: primaryColor,
+            },
+        },
+        checked: {},
```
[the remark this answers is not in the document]
MicheleK17: This is the official way of Material-ui to change the color.

src/app/Shared/Context/TableContext.tsx
```diff
@@ -2,17 +2,22 @@ import React, { useState, createContext } from "react";
 import { FilterType } from "../Model/Filter.model";
 
 export type TableType = "row" | "column";
+export type OptionType = "absolute" | "percent"
```
GerberI: consider using a more meaningful name. A option can be anything

GerberI (changes requested):
forgot to review the rest of the commit

src/app/Main/Header/Export/Header-Export.component.tsx
```diff
@@ -29,13 +29,22 @@ export function HeaderExportComponent(): JSX.Element {
...
+                    statKeys = Object.keys(table.statisticDataAbsolute[0]);
+                }
+            }
+            if (table.option === "percent") {
+                statData = table.statisticDataPercent;
+                if (!_.isEmpty(statData)) {
+                    statKeys = Object.keys(table.statisticDataPercent[0]);
+                }
+            }
+        }
```
GerberI, line 47: duplicated code could be simplified: first set statdata and then set the keys (key logic is independent of absolute or percentage data variant) => Object.keys(statData[0]).

GerberI (approved):
Commit: Display the percent values

GerberI (approved):
Commit: Only show option bar, if row/column is selected I+II
This can be simplified if the containing TableResults component does not render the options component at all if no data is present.

src/app/Main/Body/QueryPage/Results/TableResults/TableResults-Options.component.tsx
```diff
@@ -23,9 +25,20 @@ const optionsStyle = (
     flex-direction: row;
 `;
 const optionsHeadingStyle = css`
-    margin: auto 1em auto 0;
+    margin: auto 2em auto 0;
     font-weight: bold;
+    font-size: ${size}rem;
+`;
+const sizeStyle = css`
```
GerberI, line 33: consider renaming this to radioButton style or something

GerberI (approved):
Commit: Implement table option bar

GerberI (approved):
Commit: Use parseInt and handle "total = 0"
thx,
some consideration for a later refactoring or something:
It would be better if your table results data would be in the right format at the beginning. It should be parsed, validated and calculated after getting it from the API. It could look like:
interface isolateStatEntry {
name: string
stats: Record<string, number>
}
type isolateStats = isolateStatEntry[]

A results container could then map that data to the right format for the results table view component, e.g. doing the percentage calculations using the service and mapping the data to a format the results view would like to have.

PR #197, opened 2021-01-22
Use filterconfig API

MicheleK17:
- possibility to save mainFilter
    - separate selected Filter

Ticket: #108
Change-Id: Iaf83cbffc87bd5fd15a7178af97a2827913eb374

### DIFF
--- a/src/app/Core/ExportServices/generateParameterHeader.service.ts
+++ b/src/app/Core/ExportServices/generateParameterHeader.service.ts
@@ -1,7 +1,6 @@
 import { MainFilterLabelInterface } from "../../Shared/Model/Export.model";
 import {
     FilterInterface,
-    mainFilterAttributes,
 } from "../../Shared/Model/Filter.model";
 
 /**
@@ -14,7 +13,8 @@ import {
 export function generateParameterHeader(
     filter: FilterInterface,
     allFilterLabel: string,
-    mainFilterLabels: MainFilterLabelInterface
+    mainFilterLabels: MainFilterLabelInterface,
+    mainFilterAttributes: string[],
 ): string {
     const HeaderRows: string[] = [];
     HeaderRows.push("\uFEFF");
--- a/src/app/Core/ExportServices/generateParameterHeader.service.ts
+++ b/src/app/Core/ExportServices/generateParameterHeader.service.ts
@@ -20,10 +20,10 @@ export function generateParameterHeader(
     HeaderRows.push("\uFEFF");
     HeaderRows.push("#Parameter:");
 
-    mainFilterAttributes.forEach((element): void => {
+    mainFilterAttributes.forEach((element: string): void => {
         if (filter[element].length !== 0) {
             const headerFilterArray: string[] = [];
-            filter[element].forEach((filterValue) => {
+            filter[element].forEach((filterValue: string) => {
                 headerFilterArray.push(`"${filterValue}"`);
             });
             const headerFilterString = headerFilterArray.join(";");
--- a/src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
@@ -1,6 +1,7 @@
 import React, { useContext } from "react";
 import { ValueType } from "react-select";
 import { useTranslation } from "react-i18next";
+import _ from "lodash";
 import { CheckIfSingleFilterIsSet } from "../../../../../Core/FilterServices/checkIfFilterIsSet.service";
 import { DataContext } from "../../../../../Shared/Context/DataContext";
 import { FilterContext } from "../../../../../Shared/Context/FilterContext";
@@ -35,9 +36,12 @@ export function SelectorListSelectorComponent(
     const { t } = useTranslation(["QueryPage"]);
 
     const filterAttribute: FilterType = mainFilterAttributes[props.index];
-    const filterValues: string[] = filter[filterAttribute];
+    let filterValues: string[] = []
+    if (!_.isEmpty(filter.selectedFilter)) {
+        filterValues = filter.selectedFilter[filterAttribute];
+    }
     const allFilterValues: string[] = data.uniqueValues[filterAttribute];
-    const noFilter: boolean = CheckIfSingleFilterIsSet(filter, filterAttribute);
+    const noFilter: boolean = CheckIfSingleFilterIsSet(filter.selectedFilter, filterAttribute);
     return (
         <SelectorComponent
             key={`filter-selector-${filterAttribute}`}
--- a/src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
@@ -7,7 +7,6 @@ import { DataContext } from "../../../../../Shared/Context/DataContext";
 import { FilterContext } from "../../../../../Shared/Context/FilterContext";
 import {
     FilterType,
-    mainFilterAttributes,
 } from "../../../../../Shared/Model/Filter.model";
 import { TableType } from "../../../../../Shared/Context/TableContext";
 import { SelectorComponent } from "../../../../../Shared/Selector.component";
@@ -35,7 +34,7 @@ export function SelectorListSelectorComponent(
     const { filter } = useContext(FilterContext);
     const { t } = useTranslation(["QueryPage"]);
 
-    const filterAttribute: FilterType = mainFilterAttributes[props.index];
+    const filterAttribute: FilterType = filter.mainFilter[props.index];
     let filterValues: string[] = []
     if (!_.isEmpty(filter.selectedFilter)) {
         filterValues = filter.selectedFilter[filterAttribute];
--- a/src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
+++ b/src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
@@ -4,7 +4,6 @@ import { useTranslation } from "react-i18next";
 import { useContext } from "react";
 import _ from "lodash";
 import { FilterContext } from "../../../../Shared/Context/FilterContext";
-import { mainFilterAttributes } from "../../../../Shared/Model/Filter.model";
 import { ParameterContentListComponent } from "./ParameterContent-List.component";
 import { AccordionComponent } from "../../../../Shared/Accordion.component";
 
@@ -21,7 +20,7 @@ export function QueryPageParameterContentComponent(): JSX.Element {
     const { t } = useTranslation(["QueryPage"]);
 
     const displayFilter = _.cloneDeep(filter.selectedFilter);
-    mainFilterAttributes.forEach((element) => {
+    filter.mainFilter.forEach((element) => {
         if (!_.isEmpty(filter.selectedFilter) || filter.selectedFilter[element].length === 0) {
             displayFilter[element] = [t("Filters.All")];
         } else {
@@ -35,7 +34,7 @@ export function QueryPageParameterContentComponent(): JSX.Element {
      */
     const createParameterComponent = (): JSX.Element[] => {
         const elements: JSX.Element[] = [];
-        mainFilterAttributes.forEach((element): void => {
+        filter.mainFilter.forEach((element): void => {
             elements.push(
                 <ParameterContentListComponent
                     key={`parameter_list_${element}`}
--- a/src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
+++ b/src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
@@ -20,12 +20,12 @@ export function QueryPageParameterContentComponent(): JSX.Element {
     const { filter } = useContext(FilterContext);
     const { t } = useTranslation(["QueryPage"]);
 
-    const displayFilter = _.cloneDeep(filter);
+    const displayFilter = _.cloneDeep(filter.selectedFilter);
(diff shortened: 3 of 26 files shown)


src/app/Core/ExportServices/generateParameterHeader.service.ts
```diff
@@ -20,10 +20,10 @@ export function generateParameterHeader(
     HeaderRows.push("\uFEFF");
     HeaderRows.push("#Parameter:");
 
-    mainFilterAttributes.forEach((element): void => {
+    mainFilterAttributes.forEach((element: string): void => {
```
GerberI, line 23: Hint: As stated in some reviews before: forEach(element => {...}) would be perfectly fine. The correct type will automatically be inferred.

src/app/Shared/Context/FilterContext.tsx
```diff
@@ -1,19 +1,20 @@
...
-export const defaultFilter = generateEmptyFilter();
+interface FilterContextInterface {
+    mainFilter: string[];
+    selectedFilter: FilterInterface;
+    emptyFilter: FilterInterface;
+}
 
+const defaultFilter: FilterContextInterface = {
+    mainFilter: [],
+    selectedFilter: {},
```
GerberI, line 12: Consider using null if you want to express that something is explicitly not set. {} is used more for things that are yet uninitialized. Then you can also express a not set or empty state in the interface like selectedFilter: FilterInterface | null.

src/app/Shared/Context/DataContext.tsx
```diff
@@ -14,7 +13,7 @@ const defaultProfile: DataInterface = {
     ZNData: [],
     ZNDataFiltered: [],
     keyValues: [],
-    uniqueValues: defaultFilter,
+    uniqueValues: {},
```
GerberI, line 16: See FilterContext: Try to use null for expressing empty state

src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
```diff
@@ -35,9 +36,12 @@ export function SelectorListSelectorComponent(
     const { t } = useTranslation(["QueryPage"]);
 
     const filterAttribute: FilterType = mainFilterAttributes[props.index];
-    const filterValues: string[] = filter[filterAttribute];
+    let filterValues: string[] = []
+    if (!_.isEmpty(filter.selectedFilter)) {
+        filterValues = filter.selectedFilter[filterAttribute];
+    }
```
GerberI, line 41: Reject: I think you want to check if filter.selectedFilter[filterAttributes] is not undefined.

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -20,12 +20,12 @@ export function QueryPageParameterContentComponent(): JSX.Element {
     const { filter } = useContext(FilterContext);
     const { t } = useTranslation(["QueryPage"]);
 
-    const displayFilter = _.cloneDeep(filter);
+    const displayFilter = _.cloneDeep(filter.selectedFilter);
     mainFilterAttributes.forEach((element) => {
-        if (filter[element].length === 0) {
+        if (!_.isEmpty(filter.selectedFilter) || filter.selectedFilter[element].length === 0) {
```
GerberI, line 24: Reject: i think you want to check if filter.selected isEmpty and not !isEmpty
And actually i think you want to check that filter.selectedFilter[element] === undefined or filter.selectedFilter[element].length === 0

src/app/Shared/Context/FilterContext.tsx
```diff
@@ -1,19 +1,20 @@
...
-        defaultFilter[element] = [];
-    });
-    return defaultFilter;
-};
-
-export const defaultFilter = generateEmptyFilter();
+interface FilterContextInterface {
+    mainFilter: string[];
+    selectedFilter: FilterInterface;
+    emptyFilter: FilterInterface;
```
GerberI, line 7: As i see in the first commit the emptyFilter is just an initial state for selectedFilter. If so please remove it and use a constant defaultValue. If the empyFilter depends on the API response set the selectedFilter to the default value accordingly to the API response.

GerberI (commented):
Commit: Change FilterContext
General consideration
- Try to omit adding keywords like Type or Interface to the names of your interfaces and types. They provide no additional information (i learned that from Dominic). So there are language like Java where the convention is to begin an interface with an I like IFilter

src/app/Shared/Context/FilterContext.tsx
```diff
@@ -8,7 +8,7 @@ interface FilterContextInterface {
 }
 
 const defaultFilter: FilterContextInterface = {
-    mainFilter: [],
+    mainFilter: ["erreger", "matrix", "probenahmegrund"],
```
GerberI, line 11: Is this the default? If the data comes from the API i think [] is the default

GerberI (commented):
Commit: Use MainFilter from FilterContext and not model
Fine.
Consideration - there must be at least one per commit :)
- In this commit you removed one dependency of your core services from your shared model. This is in general a good thing. The core module should not depend on your other modules at all. It should be the top of your dependency hierarchy. So if a core file must depend on another  module it is a hint that the file is better placed in the module it depends from.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -3,39 +3,62 @@ import { useHistory } from "react-router-dom";
...
-import {
-    FilterContext,
-} from "../../../Shared/Context/FilterContext";
+import { FilterContext } from "../../../Shared/Context/FilterContext";
 import { TableContext } from "../../../Shared/Context/TableContext";
 import { getFilterFromPath } from "../../../Core/getFilterFromPath.service";
 import { generatePathString } from "../../../Core/generatePathString.service";
 import { getFeaturesFromPath } from "../../../Core/getTableFromPath.service";
 
+interface FilterConfigApiInterface {
```
GerberI, line 18: Consider using naming convention of our other projects. All interfaces describing data from the api should be named DTO (dataTransferObject) like FilterConfigDTO. A DTO is like a struct in higher order languages. It describes an immutable object with no functionality (no methods defined on the object).
I would also suggest to add the http mode in the name like FilterConfigResponseDTO. If you send data to the API it would be FilterConfigRequestDTO.
If you question why it is DTO and not Dto: DTO is not really a part of the name it is more a meta description and naming convention for the interface. So to recognize this on a first glance it is all upper case letters. We use this convention also for our redux actions in angular like UpdateUserSOA for StoreOperationAction.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -3,39 +3,62 @@ import { useHistory } from "react-router-dom";
...
         const r: Response = await fetch(BASE_URL);
         const dataProp: DBentry[] = await r.json();
         const keyValueProps = Object.keys(dataProp[0]) as DBkey[];
 
+        const filterResponse: Response = await fetch(FILTER_URL);
+        const filterProp: FilterConfigApiInterface = await filterResponse.json();
+
+        const mainFilter: FilterType[] = [];
         const uniqueValuesObject: FilterInterface = {};
+        const emptyFilter = {} as FilterInterface;
```
GerberI, line 46: emptyFilter: FilterInterface = {}

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -3,39 +3,62 @@ import { useHistory } from "react-router-dom";
...
+
+        filterProp.filters.forEach((filterElement) => {
+            const { name } = filterElement;
+            mainFilter.push(name);
+            uniqueValuesObject[name] = filterElement.values;
+            emptyFilter[name] = [];
```
GerberI, line 52: as stated in a previous review, here you could just init your selected filters to the default []. So no need for emptyFilter prop.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -3,39 +3,62 @@ import { useHistory } from "react-router-dom";
...
+        });
 
-        filter.mainFilter.forEach((filterElement) => {
-            const uniqueValuesPerElement: string[] = _.uniq(
-                _.map(dataProp, filterElement)
-            );
-            uniqueValuesObject[filterElement] = uniqueValuesPerElement;
+        setFilter({
```
GerberI, line 55: Is there a reason why you moved this outside the effect? Does it work if you use the browser back and forward buttons with different filter settings?

src/app/Shared/URLs.ts
```diff
@@ -7,5 +7,6 @@ export const ZNPaths = {
 };
 
 export const mockDataURL = "/v1/mockdata";
+export const filterURL = "/v1/filterconfig";
```
GerberI, line 10: In later step the constants should be provided by a config file.

GerberI (commented):
Commit: Use filter from filterconfig API
Consideration: Try to extract your API specific code to a separate sub feature: e.g. src/app/Main/API with separate model and service

GerberI (commented):
Commit: Change translation
Consider getting this info from the server to. It could provide human readable names for the attributes.

GerberI:
General consideration:
I recommend to validate the data that comes from the server (the server should validate data from the client to). This is for security reasons (e.g. if someone redirects you to his own bad server). So instead of just using filterData = parseJson(response), you should explicitly check and assign the values. So e.g. for name:
const rawData = parseJson(response)
here comes some for each i think, i omit it for simplicity
if(rawData.name === undefined) throw error
const string rawName = String.parse(rawData.name)
if(checkStringForInvalidCharacters(rawName)) throw error
filterData.name = rawName

You could put that stuff in a service function like:
parseAndValidateStringFromResponse(key: string, rawData: {}): string
and
filterData.name = parseAndValidateStringFromResponse('name', rawData)

At least you should write a ticket for data validation.

GerberI:
As you can see merging and then reviewing works. Only point is that i cannot not explicitly reject or approve the pull request anymore, but that's fine.

src/app/Shared/Context/FilterContext.tsx
GerberI: if your semantic is that selectedFilter defaults to an empty collection of filters you can forget the previous comment. But you should perhaps consider using an array or a set then.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: i moved setFilter() inside getData() because auf the API fetch. And getData() is inside the effect, so is setFilter().

GerberI: right, i didn't noticed it is inside getData

src/app/Core/ExportServices/generateParameterHeader.service.ts
MicheleK17: Done.

src/app/Shared/Context/FilterContext.tsx
MicheleK17: done. 
- Remvoce emptyFilter

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: done .

src/app/Shared/Context/FilterContext.tsx
GerberI: done too, i think

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
MicheleK17: !_.isEmpty => _.isEmpty was done with isolate api ticket

src/app/Shared/Context/DataContext.tsx
MicheleK17: if I use "undefined" as empty state, I have to check in every function "if (x !== undefined) {..} else {retrun ""}" or something like that. What would be the advantage of this? (esLint says that I should use undefined instead od null https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v20.1.0/docs/rules/no-null.md)

GerberI: I would disable this linter rule. null and undefined have different semantics. E.g. take the following api response type
user: {
id: string
name: string | null
address?: string
}
Here address is optional. So it is undefined or string. If it is undefined the data was not send. But it might be present on the server.
Name is string | null. So name = null states the data is there but it is explicitly set to null. So the user has no defined name.
So undefined is like optional and null like explicitly set to null.

So i don't know the semantics of your code here. I think uniqueValues is not optional. Perhaps the defaultFilter was the best solution. Nevertheless if you use {} as a value the compiler does not complain about that but your code can still crash if you not check if the properties of uniqueValues are set. In this case checking for null is the easier solution.

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
GerberI: and the second part? This is the same issue as stated above. If your filter.selectedFilter object is allowed to be empty, checking for non-emptiness does not imply that the filter.selectedFilter[element] is set. So the app would crash anyway. But i think you refactored your code already so that the default of selectedFilter is set to an appropriate object with empty arrays, so you can remove the isEmpty check safely. Or am i wrong?

MicheleK17: You are right. Done. 
- Remove useless _.isEmpty() condition

GerberI:
all resolved

Issue #199, opened 2021-01-25
As a User I want to have the isolates supplied by the server 


MicheleK17:
- use isolates API

PR #202, opened 2021-01-26
Use isolates form API and change model types

MicheleK17:
Ticket: #199
Change-Id: I07494d560616718c95f5005be8ca7a30eb3eca2a

### DIFF
--- a/src/app/Core/ExportServices/generateCSVString.service.ts
+++ b/src/app/Core/ExportServices/generateCSVString.service.ts
@@ -14,14 +14,19 @@ function stringModification(inputString: string): string {
  * @returns {string} - data in one csv string
  */
 export function generateCSVString<
-    T extends Record<string, string>,
+    T extends Record<string, unknown>,
     K extends keyof T
 >(dataArray: T[], headers: K[]): string {
     const csvTable: string[] = [];
 
-    dataArray.forEach((row) => {
-        const values: string[] = headers.map((header) => {
-            return stringModification(`${row[header]}`);
+    dataArray.forEach((row: T) => {
+        const values: string[] = headers.map((header: K) => {
+            const rowValue = row[header];
+            const rowValueType = typeof rowValue;
+            if (rowValueType === "string" || rowValueType === "number") {
+                return stringModification(`${rowValue}`);
+            }
+            return "";
         });
         csvTable.push(values.join(","));
     });
--- a/src/app/Core/ExportServices/generateCSVString.service.ts
+++ b/src/app/Core/ExportServices/generateCSVString.service.ts
@@ -14,7 +14,7 @@ function stringModification(inputString: string): string {
  * @returns {string} - data in one csv string
  */
 export function generateCSVString<
-    T extends Record<string, unknown>,
+    T extends Record<string, string>,
     K extends keyof T
 >(dataArray: T[], headers: K[]): string {
     const csvTable: string[] = [];
--- a/src/app/Core/ExportServices/generateCSVString.service.ts
+++ b/src/app/Core/ExportServices/generateCSVString.service.ts
@@ -22,11 +22,7 @@ export function generateCSVString<
     dataArray.forEach((row: T) => {
         const values: string[] = headers.map((header: K) => {
             const rowValue = row[header];
-            const rowValueType = typeof rowValue;
-            if (rowValueType === "string" || rowValueType === "number") {
-                return stringModification(`${rowValue}`);
-            }
-            return "";
+            return stringModification(`${rowValue}`);
         });
         csvTable.push(values.join(","));
     });
--- a/src/app/Core/ExportServices/generateFilteredDataString.service.ts
+++ b/src/app/Core/ExportServices/generateFilteredDataString.service.ts
@@ -6,14 +6,14 @@ import { generateCSVString } from "./generateCSVString.service";
  * @param {Record<string, string>[]} rawData - dataset
  * @returns {string} - dataset as string
  */
-export function RAWDataStringGenerator(
+export function generateFilteredDataString(
     rawKeys: string[],
     rawData: Record<string, string>[]
 ): string {
-    const RAWDataString: string[] = [];
+    const FilteredDataString: string[] = [];
     const headers: string[] = rawKeys;
-    RAWDataString.push(headers.join(","));
-    RAWDataString.push(generateCSVString(rawData, headers));
+    FilteredDataString.push(headers.join(","));
+    FilteredDataString.push(generateCSVString(rawData, headers));
 
-    return RAWDataString.join("\n");
+    return FilteredDataString.join("\n");
 }
--- a/src/app/Core/ExportServices/generateRAWString.service.ts
+++ b/src/app/Core/ExportServices/generateRAWString.service.ts
@@ -4,12 +4,12 @@ import { generateCSVString } from "./generateCSVString.service";
 /**
  * @desc Returns the table header and the filtered dataset as string to save it as csv
  * @param {DBkey[]} rawKeys - keys of the dataset
- * @param {DBentry[]} rawData - dataset
+ * @param {DBentry} rawData - dataset
  * @returns {string} - dataset as string
  */
 export function RAWDataStringGenerator(
     rawKeys: DBkey[],
-    rawData: DBentry[]
+    rawData: DBentry
 ): string {
     const RAWDataString: string[] = [];
     const headers: DBkey[] = rawKeys;
--- a/src/app/Core/ExportServices/generateRAWString.service.ts
+++ b/src/app/Core/ExportServices/generateRAWString.service.ts
@@ -1,18 +1,17 @@
-import { DBentry, DBkey } from "../../Shared/Model/Isolate.model";
 import { generateCSVString } from "./generateCSVString.service";
 
 /**
  * @desc Returns the table header and the filtered dataset as string to save it as csv
- * @param {DBkey[]} rawKeys - keys of the dataset
- * @param {DBentry} rawData - dataset
+ * @param {string[]} rawKeys - keys of the dataset
+ * @param {Record<string, string>[]} rawData - dataset
  * @returns {string} - dataset as string
  */
 export function RAWDataStringGenerator(
-    rawKeys: DBkey[],
-    rawData: DBentry
+    rawKeys: string[],
+    rawData: Record<string, string>[]
 ): string {
     const RAWDataString: string[] = [];
-    const headers: DBkey[] = rawKeys;
+    const headers: string[] = rawKeys;
     RAWDataString.push(headers.join(","));
     RAWDataString.push(generateCSVString(rawData, headers));
 
--- a/src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
+++ b/src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
@@ -2,10 +2,14 @@ import React, { useContext, useEffect } from "react";
 import { useHistory } from "react-router-dom";
 import _ from "lodash";
 import { DataContext } from "../../../Shared/Context/DataContext";
-import { DBentry, DBkey } from "../../../Shared/Model/Isolate.model";
-import { filterURL, mockDataURL } from "../../../Shared/URLs";
+import {
+    DBkey,
+    IsolateApiInterface,
+} from "../../../Shared/Model/Isolate.model";
+import { filterURL, isolateURL } from "../../../Shared/URLs";
 import { QueryPageComponent } from "./QueryPage.component";
 import {
+    FilterConfigApiInterface,
     FilterInterface,
     FilterType,
 } from "../../../Shared/Model/Filter.model";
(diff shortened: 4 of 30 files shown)


src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -15,28 +19,19 @@ import { getFilterFromPath } from "../../../Core/getFilterFromPath.service";
...
-}
-
 export function QueryPageContainerComponent(): JSX.Element {
     const { data, setData } = useContext(DataContext);
     const { filter, setFilter } = useContext(FilterContext);
     const { table, setTable } = useContext(TableContext);
     const history = useHistory();
 
-    const BASE_URL: string = mockDataURL;
+    const BASE_URL: string = isolateURL;
```
GerberI, line 28: ISOLATE_URL?

src/app/Shared/Model/Isolate.model.ts
```diff
@@ -1,3 +1,175 @@
...
+                value: string;
+                active: boolean;
+            };
+        };
+    }[];
+}
 
-export type DBentry = Record<DBkey, string>;
+
+export type DBentry = IsolateApiInterface["isolates"];
```
GerberI: Reject: your business logic model should not depend on the api model,  i think your actual app does not interpret all the stuff in the api model? So keep an independent model for your app logic and map the api model to that model (only the data that is necessary and to a format that fits your components inputs)

GerberI: Reject2: your dbentry is no entry anymore, it is a collection of the dbentries, so you should rename it

src/app/Shared/Model/Isolate.model.ts
```diff
@@ -1,3 +1,175 @@
...
+
+export type DBkey = keyof IsolateApiInterface["isolates"][0];
```
GerberI: Reject: see dbentry

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -15,28 +19,19 @@ import { getFilterFromPath } from "../../../Core/getFilterFromPath.service";
...
     const FILTER_URL: string = filterURL;
 
     const getData = async (): Promise<void> => {
-        const r: Response = await fetch(BASE_URL);
-        const dataProp: DBentry[] = await r.json();
-        const keyValueProps = Object.keys(dataProp[0]) as DBkey[];
+        const isolateResponse: Response = await fetch(BASE_URL);
+        const isolateProp: IsolateApiInterface = await isolateResponse.json();
+        const keyValueProps = Object.keys(isolateProp.isolates[0]) as DBkey[];
```
GerberI, line 34: As stated in a previous review, consider refactoring this. Your data should not depend on the first entry of an api response. The client should know what it expects from the api and the data format should be clear for both sides, client and server.

src/app/Core/ExportServices/generateCSVString.service.ts
```diff
@@ -14,14 +14,19 @@ function stringModification(inputString: string): string {
  * @returns {string} - data in one csv string
  */
 export function generateCSVString<
-    T extends Record<string, string>,
+    T extends Record<string, unknown>,
```
GerberI: why is this unknown? Your code should be aware of how the data looks like it processes

GerberI (changes requested):
Commit: Use isolates form API and change model types
In general the data format of the API does not look very flat. I recommend to keep the API as simple, flat and anambiguously as possible.

src/app/Core/ExportServices/generateRAWString.service.ts
```diff
@@ -1,18 +1,17 @@
...
- * @param {DBentry} rawData - dataset
+ * @param {string[]} rawKeys - keys of the dataset
+ * @param {Record<string, string>[]} rawData - dataset
  * @returns {string} - dataset as string
  */
 export function RAWDataStringGenerator(
-    rawKeys: DBkey[],
-    rawData: DBentry
+    rawKeys: string[],
+    rawData: Record<string, string>[]
```
GerberI: nice. Actually the stringGenerator is independent of the model.

src/app/Core/ExportServices/generateCSVString.service.ts
MicheleK17: Already changed it in the commit: "Adaption to isolate API "

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -36,6 +36,13 @@ export function QueryPageContainerComponent(): JSX.Element {
         const filterResponse: Response = await fetch(FILTER_URL);
         const filterProp: FilterConfigApiInterface = await filterResponse.json();
 
+        filterProp.filters.forEach((element, index) => {
+            if (element.id === "sContext") {
+                filterProp.filters[index].id = "samplingContext";
+                filterProp.filters[index].name = "samplingContext";
+            }
+        });
```
GerberI, line 44: Reject: It is in general a bad idea to manipulating the data structure you are iterating over. I would suggest to map the filterProps to the desired datastructure. So const newFilterProps =  filterProp.filters.map(...)...

src/app/Shared/Model/Isolate.model.ts
```diff
@@ -165,7 +165,7 @@ export interface IsolateApiInterface {
                 value: string;
                 active: boolean;
             };
-        };
+        }; */
```
GerberI, line 168: See comment in review of previous commit

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
GerberI: See comment in previous commit: Here you already mapping the api data to a more appropriate structure fitting your components desire. So you should use separate models for api and client data.

GerberI (changes requested):
Commit: Use isolates form API and change model types

src/app/Core/ExportServices/generateFilteredDataString.service.ts
```diff
@@ -6,14 +6,14 @@ import { generateCSVString } from "./generateCSVString.service";
  * @param {Record<string, string>[]} rawData - dataset
  * @returns {string} - dataset as string
  */
-export function RAWDataStringGenerator(
+export function generateFilteredDataString(
```
GerberI, line 9: not so good: now you renamed the universally usable function so it suggest to be only work with filteredData, but the name is better than the previous one

GerberI (approved):
Renaming inconsistent components
Some renaming is always good :)

src/app/Core/ExportServices/generateCSVString.service.ts
GerberI: then your typechecking code below is also obsolete?

GerberI (approved):
Commit: CR filterconfig-API
thx

GerberI (approved):
Commit: Renaming
thx,
i predict that this was not the last renaming of the Db... types :)

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: done. 
- 1. Reject in "Only use desired values from isolate-API"
- 2.Comment in "Separate api_isolate_model from data_isolate_model"

src/app/Core/ExportServices/generateCSVString.service.ts
MicheleK17: done. 
- Remove TypeCheck in generateCSVString

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: Done. 
- Only use desired values from isolate-API

src/app/Shared/Model/Isolate.model.ts
MicheleK17: done.

src/app/Shared/Model/Isolate.model.ts
MicheleK17: done. 
- Only use desired values from isolate-API

src/app/Shared/Model/Isolate.model.ts
MicheleK17: done.

    Only use desired values from isolate-API

src/app/Shared/Model/Isolate.model.ts
```diff
@@ -165,11 +164,16 @@ export interface IsolateDTO {
                 value: string;
                 active: boolean;
             };
-        }; */
+        };
```
GerberI: Reject? Are these API changes?
The DTO should exactly reflect the data you get from the API.

src/app/Shared/Model/Isolate.model.ts
```diff
@@ -165,11 +164,16 @@ export interface IsolateDTO {
...
+    microorganism: string;
+    samplingContext: string;
+    matrix: string;
+}[];
 
-export type DbCollection = IsolateDTO["isolates"];
+export type DbKey = "microorganism" | "samplingContext" | "matrix";
 
-export type DbKey = keyof IsolateDTO["isolates"][0];
+export const DbValues: DbKey[] = ["microorganism", "samplingContext", "matrix"];
```
GerberI: DbKeys would be a better name

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -31,23 +32,32 @@ export function QueryPageContainerComponent(): JSX.Element {
...
         const isolateProp: IsolateDTO = await isolateResponse.json();
-        const keyValueProps = Object.keys(isolateProp.isolates[0]) as DbKey[];
+
+        const adaptedDbIsolates: DbCollection = isolateProp.isolates.map(
+            ({ microorganism, samplingContext, matrix }) => ({
+                microorganism,
+                samplingContext,
+                matrix,
+            })
+        );
```
GerberI, line 42: Nice. Now if the api keys changes you just need to change this map function and the rest of your code works as before
Same for the filter props below

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -31,23 +32,32 @@ export function QueryPageContainerComponent(): JSX.Element {
...
 
         const filterResponse: Response = await fetch(FILTER_URL);
         const filterProp: FilterConfigDTO = await filterResponse.json();
 
-        filterProp.filters.forEach((element, index) => {
-            if (element.id === "sContext") {
-                filterProp.filters[index].id = "samplingContext";
-                filterProp.filters[index].name = "samplingContext";
+        const adaptedFilterProp = filterProp.filters.map(filterObj => {
+            const tempFilterProp = { ...filterObj};
```
GerberI: Consider using a more safe way to map the data. So explicitly create the new objects:
tempFilterProp = {
id: filterObj.id
name: filterObj.name
...
}
Another way would be using a json validator/schema during the json parse step.
Reason: Actually you don't know what the fetched data looks like. By typing it to the DTO you only tell the compiler how the data should look like. At runtime invalid data will be also mapped to your adaptedFilterProps and your app might crash somewhere else in the code. By explicitly mapping the values here your app will throw an exception also at his place, so you now whats going wrong.

GerberI (changes requested):
Commit: Only use desired values from isolate-API
Much better now

GerberI (approved):
Commit: Separate api_isolate_model from data_isolate_model
I would move the api specific code to a separate sub feature. So then you don't need to name your model files client_model. In general your actual business logic things should have no amendments to their names. So API_model or IsolateDTO is fine because they are not part of your business logic.

GerberI (approved):
Commit: Remove TypeCheck in generateCSVString

src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -6,4 +6,4 @@ export type DbCollection = {
 
 export type DbKey = "microorganism" | "samplingContext" | "matrix";
 
-export const DbValues: DbKey[] = ["microorganism", "samplingContext", "matrix"];
+export const DbKeyCollection: DbKey[] = ["microorganism", "samplingContext", "matrix"];
```
GerberI, line 9: 👍 you can ignore my previous DbKeys suggestion

GerberI (approved):
Commit: Remove emptyFilter
nice

GerberI (approved):
Commit: Remove unused isTable-prop in Option branch
I almost forgot about that (but i did not :))
thx

src/app/Shared/Model/Isolate.model.ts
MicheleK17: As it is now, it is the current api.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: done. 
- Change how to adapted FilterProp

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -13,13 +13,27 @@ import {
...
     FilterType,
+    SingleFilterConfig,
 } from "../../../Shared/Model/Filter.model";
 import { FilterContext } from "../../../Shared/Context/FilterContext";
 import { TableContext } from "../../../Shared/Context/TableContext";
 import { getFilterFromPath } from "../../../Core/getFilterFromPath.service";
 import { generatePathString } from "../../../Core/generatePathString.service";
 import { getFeaturesFromPath } from "../../../Core/getTableFromPath.service";
 
+function setAdaptedFilterProp(
```
GerberI, line 24: getAdaptedFilterProp, but you can leave as is because it is a private function

GerberI (approved):
Commit: Change how to adapted FilterProp
Good, you also avoided duplicated code 👍

PR #206, opened 2021-02-01
Change label of selectors for displayed features

MicheleK17:
Ticket: #204
Change-Id: I242df01d6ddd9d0b94433e18f0e1ce01cdad8247

### DIFF
--- a/src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
@@ -49,6 +49,7 @@ export function DisplayedFeatureSelectorComponent(
             selectedValues={selectedValues}
             isMulti={false}
             isNotSelect={isNotSelect}
+            isFeature
         />
     );
 }
--- a/src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
@@ -12,14 +12,10 @@ function generateTranslatedSelectorObject(
     selectorArray: string[],
     translateFunction: TFunction
 ): { value: string; label: string }[] {
-    const selectorLabelObject: {
-        value: string;
-        label: string;
-    }[] = selectorArray.map((selectorElement: string) => {
+    return selectorArray.map((selectorElement: string) => {
         const label: string = translateFunction(`Filters.${selectorElement}`);
         return { value: selectorElement, label };
     });
-    return selectorLabelObject;
 }
 export interface FeatureSelectorProps {
     activeFeature: FilterType;
--- a/src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
@@ -1,13 +1,26 @@
 import React from "react";
 import { ValueType } from "react-select";
+import { useTranslation } from "react-i18next";
+import { TFunction } from "i18next";
 import _ from "lodash";
 import { SelectorComponent } from "../../../../../Shared/Selector.component";
 import { TableType } from "../../../../../Shared/Context/TableContext";
-import {
-    FilterType,
-} from "../../../../../Shared/Model/Filter.model";
+import { FilterType } from "../../../../../Shared/Model/Filter.model";
 import { generateFeatureList } from "../../../../../Core/generateFeatureList.service";
 
+function generateTranslatedSelectorObject(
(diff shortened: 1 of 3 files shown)


GerberI (approved):
I would recommend a little refactoring. As your SelectorComponent is a common shared component i would move the code interacting with the translation framework up the hierarchy to the feature specific components using the SelectorComponent. So the SelectorComponent .e.g. has a prop labels. The SelectorComponent does not need to know about the isFeature.
A quicker but not so clean solution would be to use a prop useTranslation: string | false, it would be set to 'QueryPage' then.

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
```diff
@@ -1,13 +1,26 @@
...
+): { value: string; label: string }[] {
+    const selectorLabelObject: {
+        value: string;
+        label: string;
+    }[] = selectorArray.map((selectorElement: string) => {
+        const label: string = translateFunction(`Filters.${selectorElement}`);
+        return { value: selectorElement, label };
+    });
+    return selectorLabelObject;
+}
```
GerberI, line 23: can be simplified: return selectorArray.map(...)

src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
```diff
@@ -4,12 +4,23 @@ import { useTranslation } from "react-i18next";
...
+    const selectorLabelObject: {
+        value: string;
+        label: string;
+    }[] = selectorArray.map((filterValue) => {
+        return { value: filterValue, label: filterValue };
+    });
+
+    return selectorLabelObject;
+}
+
```
GerberI, line 23: same simplification as above: return selectorArray.map(...)

src/app/Shared/Selector.component.tsx
```diff
@@ -75,68 +73,32 @@ const selectStyle: StylesConfig = {
...
-        selectorObject.push({ value: element, label });
-    });
-    return selectorObject;
-}
-
 export interface SelectorProps {
     label: string;
-    selectValues: string[];
+    dropDownValuesObj: { value: string; label: string }[];
+    selectedValuesObj: { value: string; label: string }[];
```
GerberI, line 79: question: is the selectedValues a slice of the data of the dropDownValues? If so it would be sufficient to give the selectedValues as an value[] where the values refer to the values in the dropDownValuesObj.

GerberI (approved):
Commit: Refactoring & Renaming
thx,
Some more words to the motivation of this refactoring:
In general you try to code your view components or reusable components as clear as possible. So you move your 'messy' code to the container components or up the component hierarchy. So now your generatedTranslatedSelectorObj is duplicated, so a bit messier but that's okay, you traded it off to keep your shared SelectorComponent clean. So if the duplicate code becomes more complex you always have the option to put it into a service or helper function.

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
MicheleK17: done .

src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
MicheleK17: done .

src/app/Shared/Selector.component.tsx
MicheleK17: Yes they are. But the selector package expects an ObjectList also for selectedValues.

GerberI: Ok, if you feel like, you can move the selecetedValuesObj generating code here, so that the selectedValuesProps is only a value[]. But perhaps sometimes you want the selected labels to be different than the initial labels, so you can also leave it as it is.

GerberI (approved):
Commit: Simplified: return selectorArray.map(...)
So nothing to say, so another consideration :)
I often see that you use something like:
const obj: { foo: string, bar: string } = someFunction();
So because we developers are lazy you can also facilitate type inference here:
const obj = someFunction();
The obj will be of the correct type. This makes it also easier to change the function return type. Then you have to change the type of obj only at the place where it is actually used.
Only exception i make is when i define a new object like:
const obj = {
foo: string,
bar: string
}
Here i prefer to type the obj to make the definition more safe:
const obj: {foo: string, bar: string} = {
foo: string,
bar: string
}
So as reminder, in general you state the types only when you declare something, like a function header, an interface or a new object.

src/app/Shared/Selector.component.tsx
MicheleK17: I think it's okay the way it is for now.

GerberI:
ready for merge

Issue #203, opened 2021-01-27
As a user, I want to be informed when the server is not responding with data or the URL is not correct.


MicheleK17:
Change "loading data ..." to a more meaningful message when no data is given by the server

MicheleK17:
Change "loading data ..." to a more meaningful message when no data is given by the server

dptolle:
Tested locally

PR #207, opened 2021-02-01
Implement ErrorPage and Loading sign

MicheleK17:
Ticket: #203
Change-Id: I595d969111b86840c0a59e8e95deb4f6214899ab


src/app/Main/Body/QueryPage/Container-ChooseComponent.component.tsx
```diff
@@ -0,0 +1,27 @@
...
+            return <ContainerPageNotFoundComponent />;
+        }
+        if (
+            props.dataIsSet === true &&
+            props.status.isolateStatus === 200 &&
+            props.status.filterStatus === 200
+        ) {
+            return <QueryPageComponent />;
+        }
+    }
```
GerberI: Reject: and if the status is something else? Then this component does nothing.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -114,10 +131,7 @@ export function QueryPageContainerComponent(): JSX.Element {
...
     }, [filter, table]);
 
-    let returnValue = <p> Loading data ... </p>;
-    if (_.isEmpty(data.ZNData) === false) {
-        returnValue = <QueryPageComponent />;
-    }
-
-    return returnValue;
+    return (
+        <ChooseComponent status={status} dataIsSet={!_.isEmpty(data.ZNData)} />
```
GerberI: Reject: the status already expresses if data was set, so dataIsSet is obsolete. You also check in this component for status code 200 and in ChooseComponent again. So you should decide which component is responsible for checking the status code and do it only once.
So in this state of the code (first commit) i would check the status here and remove the status property from ChooseComponent. The ChooseComponent does not need to know the status codes and which http request failed. It only needs to know if the app is loading, loading failed, or loading succeded.

src/locales/de/ErrorPage.json
```diff
@@ -0,0 +1,4 @@
+{
+    "Subtitle": "Seite nicht gefunden",
+    "MainText": "Sie haben unsere Fehlerseite gefunden. Dies ist nicht Ihre Schuld, sondern ein Fehler auf unserer Serverseite. Bitte entschuldigen Sie die entstandenen Unannehmlichkeiten und kontaktieren Sie gern unser Team unter dem Reiter 'Probleme melden'."
+}
```
GerberI, line 4: I would rewrite this. A simple "Die Seite konnte nicht geladen werden" would be fine. We are not always responsible for network errors. And the user might also be responsible for the error. And the user is not the president of the united states who might get angry about "Unannehmlichkeiten"

GerberI (changes requested):
Commit: Implement ErrorPage and Loading sign

src/app/Shared/LogoCard.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    {props.title}
+                </Typography>
+                <Typography variant="h2" css={subtitleStyle}>
+                    {props.subtitle}
+                </Typography>
+                <Typography css={textStyle}>{props.text}</Typography>
+            </CardContent>
+            <CardMedia
+                component="img"
+                alt="Contemplative Reptile"
```
GerberI, line 71: ReReject: rename this please

GerberI (changes requested):
Implement ErrorPage and Loading sign
nice,
As a reminder: Now your ContainerPageNotFoundComponent and HomePageLaoutComponent are typically container components. They fetch the data (here the translation) and contain a simple view component (the LogoCard). So container contains the logic and the view is stupid as it should be.

src/app/Main/Body/ErrorPage/PageNotFound-TextContent.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+const normalTextStyle = css`
+    font-size: 0.85rem;
+    line-height: 1.6;
+    text-align: justify;
+`;
+
+export function ErrorPageTextContentComponent(): JSX.Element {
+    const { t } = useTranslation(["ErrorPage"]);
+    return (
+        <CardContent>
```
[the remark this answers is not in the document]
GerberI: obsolete due to next commit

src/app/Main/Body/ErrorPage/Container-PageNotFound.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export function ContainerPageNotFoundComponent(): JSX.Element {
+    const classes = useStyles();
+
+    return (
+        <Card className={classes.root}>
+            <ErrorPageTextContentComponent />
+            <CardMedia
+                component="img"
+                alt="Contemplative Reptile"
```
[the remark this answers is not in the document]
GerberI: rejected in the next commit

src/locales/de/ErrorPage.json
GerberI: obsolete

src/locales/de/ErrorPage.json
```diff
@@ -1,4 +1,4 @@
 {
     "Subtitle": "Seite nicht gefunden",
-    "MainText": "Sie haben unsere Fehlerseite gefunden. Dies ist nicht Ihre Schuld, sondern ein Fehler auf unserer Serverseite. Bitte entschuldigen Sie die entstandenen Unannehmlichkeiten und kontaktieren Sie gern unser Team unter dem Reiter 'Probleme melden'."
+    "MainText": "Diese Seite konnte nicht gefunden werden. Vielleicht besteht ein Fehler in der eingegebenen URL oder auf unserer Serverseite. Bitte entschuldigen Sie die entstandenen Unannehmlichkeiten und kontaktieren Sie gern unser Team unter dem Reiter 'Probleme melden'."
```
GerberI, line 3: I would remove the "Bitte entschuldigen ... Unannehmlichleiten" part too

src/app/Main/CheckServerConnection.component.tsx
```diff
@@ -25,7 +26,7 @@ export function CheckServerConnectionComponent(): JSX.Element {
     });
 
     if (loading) {
-        return <p> </p>;
+        return <LoadingProcessComponent />;
```
GerberI, line 29: Question: How does the CheckServerConnectionComponent work? Where it is used and when the connection is checked? And why the connection is checked?

GerberI (changes requested):
Commit: Implement ErrorPage and Loading sign
good

src/app/Main/Body/HomePage/HomePage.component.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import { useTranslation } from "react-i18next";
 import { LogoCardComponent } from "../../../Shared/LogoCard.component";
 
-export function HomePageLayoutComponent(): JSX.Element {
+export function HomePageComponent(): JSX.Element {
```
GerberI, line 5: good, i already thought to reject the name in a previous commit

src/app/Main/Body/QueryPage/Container-Choose.component.tsx
```diff
@@ -3,15 +3,20 @@ import { ErrorPagePageNotFoundComponent } from "../ErrorPage/ErrorPage-PageNotFo
...
 import { LoadingProcessComponent } from "../../../Shared/LoadingProcess.component";
 
-export function ChooseComponent(props: {
-    status: {
-        isolateStatus: number;
-        filterStatus: number;
-    } | undefined;
+export function ContainerChooseComponent(props: {
+    status:
+        | {
```
GerberI, line 14: ? Why is there a pipe | before the {

GerberI (changes requested):
Commit: Rename and use "Prettier"

src/app/Main/Body/QueryPage/Container-Choose.component.tsx
```diff
@@ -13,11 +19,19 @@ export function ContainerChooseComponent(props: {
...
+                    errorStatus={props.status.isolateStatus}
+                />
+            );
+        }
+        if (props.status.filterStatus !== 200) {
+            return (
+                <ErrorPagePageNotFoundComponent
+                    errorStatus={props.status.filterStatus}
+                />
+            );
```
GerberI, line 34: You can remove the 200 checks in the next statement now. There are obsolete

src/app/Shared/LogoCard.component.tsx
```diff
@@ -45,6 +45,13 @@ export interface CardProps {
     text: string;
 }
 
+/**
+ * @desc Returns an card wrapper with BfR-Logo
```
GerberI, line 49: a card wrapper

GerberI (changes requested):
Commit: Add function-desc. & make ErrorPage universal
General considerations:
- I would separate page not found error messages from fetch error messages. So if the fetch fails the frontend could show a message like server connection failed or so. This is because fetching data from the server is not the same as navigating to an unknown page.
- I would kind of separate the initial data fetching to a separate service. So you should differentiate the data that is necessary to make the app working and the data that is fetched during the usage of the app. So if the page is loaded and you fetch data during using the app displaying a page with 'page not found' is not very user friendly, here a short message informing the user about a connection failure is more appropiate. The other way, if the app is not able to load the data it needs to work at all, showing an error page is the better way.

src/app/Main/Body/QueryPage/Container-ChooseComponent.component.tsx
GerberI: obsolete

src/app/Main/Body/QueryPage/Container-Choose.component.tsx
MicheleK17: I don't know - it came with the "prettier - code formatter"

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: 1. No it's not obsolete. React always render twice. If I do not check if the Data is already set, I get the error “Rendered more hooks than during the previous render” in the QueryPage.

src/app/Shared/LogoCard.component.tsx
MicheleK17: done .

src/locales/de/ErrorPage.json
MicheleK17: It is only a text proposal. Katja (and Dominic) will change it again anyway.

src/app/Main/CheckServerConnection.component.tsx
MicheleK17: This check was written as part of another ticket. It checks if there is in general a connection to the server (/v1). If not, it is not possible to access the query page at all (tab is blocked).

src/app/Main/Body/ErrorPage/ErrorPage-PageNotFound.component.tsx
```diff
@@ -2,12 +2,21 @@ import React from "react";
...
+ * @desc Return an ErrorPage card with the current error status
+ * @param {number} errorStatus - status of the api fetch
+ * @returns {JSX.Element} - card with ErrorPage content
+ */
+export function ErrorPagePageNotFoundComponent(props: {
+    errorStatus: number;
+}): JSX.Element {
     const { t } = useTranslation(["ErrorPage"]);
 
+    const errorSting = props.errorStatus.toString();
```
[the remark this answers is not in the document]
MicheleK17: done .

src/app/Shared/LogoCard.component.tsx
MicheleK17: done .

src/app/Main/Body/QueryPage/Container-ChooseComponent.component.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import { ContainerPageNotFoundComponent } from "../ErrorPage/Container-PageNotFound.component";
+import { ContainerLoadingProcessComponent } from "./Container-LoadingProcess.component";
+import { QueryPageComponent } from "./QueryPage.component";
+
+export function ChooseComponent(props: {
```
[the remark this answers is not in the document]
MicheleK17: I know. I tried a new name, but I still don't like it very much.

src/app/Main/Body/QueryPage/Container-Choose.component.tsx
GerberI: ok, this is allowed in typescript. The formatter makes this so that types like:
type foo =
| foo1
| foo2
| foo3
...
looks prettier. It's the same like:
const foo = {
foo1,
foo2,
foo3, (here is a comma at the end)
}
I'm not a fan of that 'convention' because it looks like if someone forgot a type or property.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
GerberI: Ok, a similar error does exist in angular. It is check if your code is correctly designed. And if you get that message it is not. I would suggest to read some documentation about the message and try to understand why this error happens and where in your code is the problem. I'm not into this, but i would guess that you define the data hook only if the connection is a success, so your hooks change during render cycles. The solution might be to put the data hook in a child component that renders only if the connection succeeded. So in general the hooks should stick to the components, if the component is ready the hook should be too.

GerberI (approved):
Commit: Renaming
thx

src/app/Main/Body/QueryPage/Container-ChooseComponent.component.tsx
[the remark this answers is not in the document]
GerberI: Perhaps QueryPageLoadingWrapper or something? In general you do such stuff in your top most app component instead of an own component.

src/app/Main/CheckServerConnection.component.tsx
GerberI: Hm, ok. As stated in another comment, i think it would be better to show an error message if the connection gets lost. The initial server connection is already checked by the former 'ChooseComponent'.
Another question: When is this checked? On page load, user interaction or in a time intervall?

MicheleK17: It was Dominic's idea to do it that way. I have to discuss it with him - if I want to do it differently. The check is done once on page load.

GerberI: Ok, discussion is always good :). So if it is done on page load it seems obsolete. If you discuss this with dominic think of the type of data that you use in your app. What data is essential to get the app working and is loaded at page load and what data is fetched by user interaction. Best would be if there is only one point in the app handling the essential data loading. And as stated, the data fetching should be handled on the fly by displaying error messages (in my opinion)

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: Will be a future ToDo.

src/app/Main/Body/ErrorPage/ErrorPage-PageNotFound.component.tsx
```diff
@@ -2,12 +2,21 @@ import React from "react";
...
 import { LogoCardComponent } from "../../../Shared/LogoCard.component";
 
-export function ErrorPagePageNotFoundComponent(): JSX.Element {
+/**
+ * @desc Return an ErrorPage card with the current error status
+ * @param {number} errorStatus - status of the api fetch
+ * @returns {JSX.Element} - card with ErrorPage content
+ */
+export function ErrorPagePageNotFoundComponent(props: {
+    errorStatus: number;
```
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Main/Body/QueryPage/Container-Choose.component.tsx
MicheleK17: Done.

src/app/Main/Body/QueryPage/Container-ChooseComponent.component.tsx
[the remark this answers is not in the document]
MicheleK17: I'll leave the name as it is for now. Maybe everything will be changed again anyway, when I deal with the hooks-problem again.

src/app/Main/Body/QueryPage/QueryPage-LoadingOrError.component.tsx
```diff
@@ -21,23 +21,15 @@ export function QueryPageLoadingOrErrorComponent(props: {
...
-                />
+                <ErrorPageComponent errorStatus={props.status.isolateStatus} />
             );
         }
         if (props.status.filterStatus !== 200) {
             return (
-                <ErrorPageComponent
-                    errorStatus={props.status.filterStatus}
-                />
+                <ErrorPageComponent errorStatus={props.status.filterStatus} />
```
GerberI, line 29: General idea for a later refactoring:
I would suggest to use some function: 
fetchQueryPageData(): Status {
status = fetchData1()
if status === 404 return Status.PageNotFound;
if status !== 200 return Status.GeneralError;
status = fetchData2()
if status === 404 ...
if status === 200 ...
return Status.Success
}
And in the QueryPage container compoment:
let status = Status.Loading;
status = fetchQueryPageData(); // this is in some async hook
if(loading) return LoadingComponent
if(status === Status.Success) return QueryPage
else return ErrorPage(status)
Idea is that there is only one status independent of how many request are needed for the fetching all the data. So you don't have to differentiate between isolateStatus and filterStatus. And the actually http status code handling is separated from the rest of your code and at the same place where the fetching happens.

src/locales/de/ErrorPage.json
```diff
@@ -1,4 +1,7 @@
 {
-    "Subtitle": "Seite nicht gefunden",
-    "MainText": "Diese Seite konnte nicht gefunden werden. Vielleicht besteht ein Fehler in der eingegebenen URL oder auf unserer Serverseite. Bitte entschuldigen Sie die entstandenen Unannehmlichkeiten und kontaktieren Sie gern unser Team unter dem Reiter 'Probleme melden'."
+    "PageNotFoundTitle": "Seite nicht gefunden",
+    "PageNotFoundText": "Diese Seite konnte nicht gefunden werden. Vielleicht besteht ein Fehler in der eingegebenen URL oder auf unserer Serverseite. Gern können Sie  unser Team unter dem Reiter 'Probleme melden' konatktieren.",
+    "GeneralErrorTitle": "Ein Fehler ist aufgetreten.",
+    "GeneralErrorText": "Vielleicht besteht ein Fehler auf unserer Serverseite. Gern können Sie  unser Team unter dem Reiter 'Probleme melden' konatktieren."
```
GerberI, line 5: typo in both messages: konatkieren.
But you said that the messages are only a placeholder, so no reject

GerberI (commented):
Commit: Remove 200-check & change ErrorText
thx

GerberI:
all resolved

Issue #208, opened 2021-02-02
Get filter from URL-Path by use defined MainFilter.


MicheleK17:

dptolle:
@MicheleK17 What is this bug?

MicheleK17:
Bug: It was no longer possible to read the filters from the URL.
Reason: Within the Filter api ticket I moved the setting of the filters to another function.

PR #211, opened 2021-02-04
Debug loading Filter from URL

MicheleK17:
Ticket: #208
Change-Id: Id9855b267b3a9d8940eb945cf2d4530321242b45


GerberI (approved):
So you removed fetching the mainFilters from the api in this ticket? If so, i would change the ticket name accordingly.

Issue #214, opened 2021-02-08
Switch from processing the mock-data in the front end to utilizing the server API for displaying and downloading the derived data set


dptolle:
endpoint: /insolat/counted

PR #215, opened 2021-02-10
Use isolate/counted api for: data table, nr of isolates and file export

MicheleK17:
- use sContext instead of samplingContext

Ticket: #214
Change-Id: Ie230455e5ed30da8893984de003b36cb545124a0


GerberI (approved):
Commit: Adapt isolate from api to sContext

~Why you renamed your internal data structures to sContext? We discussed it would be good to be independent from api and map the data to internal structures at one point in the code (some api service). Now you have a cryptic identifier sContext only because it is shorter in the url. This should not be a reason to make the code harder to understand.~
So the api now uses also samplingContext?

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -48,11 +52,12 @@ export function QueryPageContainerComponent(): JSX.Element {
         if (isolateStatus === 200 && filterStatus === 200) {
```
GerberI: Reject: && isolateCountStatus === 200

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -64,15 +69,28 @@ export function QueryPageContainerComponent(): JSX.Element {
                 uniqueValuesObject[id] = filterElement.values;
             });
 
+            const nrOfSelectedIsolates = isolateCountProp.totalNumberOfIsolates;
+
             setData({
+                ...data,
```
GerberI, line 87: ? is there some other data that is not setted here?

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -64,15 +69,28 @@ export function QueryPageContainerComponent(): JSX.Element {
...
                 ZNData: adaptedDbIsolates,
                 ZNDataFiltered: adaptedDbIsolates,
                 keyValues: DbKeyCollection,
                 uniqueValues: uniqueValuesObject,
+                nrOfSelectedIsolates,
             });
         }
     };
 
+    const fetchIsolateCounted = async (): Promise<void> => {
```
GerberI: duplicated code. You could use this function above

src/app/Shared/Context/DataContext.tsx
```diff
@@ -7,13 +7,15 @@ interface DataInterface {
     ZNDataFiltered: DbCollection;
     keyValues: DbKey[];
     uniqueValues: FilterInterface;
+    nrOfSelectedIsolates: number;
```
GerberI, line 9: Is this always the same as ZNDataFiltered.length?

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -31,11 +33,13 @@ export function QueryPageContainerComponent(): JSX.Element {
     const history = useHistory();
 
     const ISOLATE_URL: string = isolateURL;
+    const ISOLATE_COUNT_URL: string = isolateCountURL + history.location.search;
```
GerberI, line 48: consider removing upper case style because these are not constants anymore

src/app/Shared/URLs.ts
```diff
@@ -9,5 +9,6 @@ export const ZNPaths = {
 export const mockDataURL = "/v1/mockdata";
 export const filterURL = "/v1/filterconfig";
 export const isolateURL = "/v1/isolate";
+export const isolateCountURL = "/v1/isolate/counted/";
```
GerberI, line 12: these are constants and should be upper case style

GerberI (changes requested):
Commit: Use nrOfSelectedIsolates from counted api
Does the counted endpoint provide some data that the filtered data endpoint does not provide or that is not derivable from that data? If not why is there a counted endpoint at all?

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -1,126 +1,125 @@
...
+                ? [t("Results.TableHead")]
+                : (_.isEmpty(filter.selectedFilter[rowAttribute])
+                ? data.uniqueValues[rowAttribute]
+                : filter.selectedFilter[rowAttribute])
+
+            const colValues: string[] = !isCol
+                ? [t("Results.TableHead")]
+                : (_.isEmpty(filter.selectedFilter[colAttribute])
+                ? data.uniqueValues[colAttribute]
+                : filter.selectedFilter[colAttribute])
```
GerberI: try to avoid nested ? : operators. They should only be used for short and simple statements. So for readability i would prefer something like:
rowValues = t("...")
if(isRow) {
const isEmpty = _.isEmpty(...)
rowValues = isEmpty ? data.uniqueValues : filter.selectedValues
}

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -1,126 +1,125 @@
...
-            const rowsWithIsolates = generateRowsWithIsolates(
-                dataToCountIsolates,
-                colAttribute,
-                rowAttribute,
-                rowValues,
-                [t("Results.TableHead")],
-                "row"
+
+            const nrOfSelectedIsolates = isolateCountProp.totalNumberOfIsolates;
+            const rowsWithPercent = calculateRowsWithPercent(
```
GerberI: consider rename to statisticTableDataRel or something

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -1,126 +1,125 @@
...
+
+            const colValues: string[] = !isCol
+                ? [t("Results.TableHead")]
+                : (_.isEmpty(filter.selectedFilter[colAttribute])
+                ? data.uniqueValues[colAttribute]
+                : filter.selectedFilter[colAttribute])
+
+            const isolateColValues: string[] = [t("Results.TableHead")];
+            const isolateRowValues: string[] = [t("Results.TableHead")];
+            isolateCountProp.groups.forEach((isolateCount) => {
```
GerberI: i think this whole algorithm can be simplified. E.g. you iterate over isolateCountGroups here and again for each rowValue and colValue pair below. I don't exactly what the algorithm does (and have no time to get into it), so i can give no tip how to make it better. But if you have the time you can refactor it. Or we can discuss this over jitsi, so you can explain what the algo should do.

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -1,126 +1,125 @@
...
 
+    const ISOLATE_COUNT_URL: string = isolateCountURL + history.location.search;
+
+    const fetchIsolateCounted = async (): Promise<void> => {
+        const isolateCountResponse: Response = await fetch(ISOLATE_COUNT_URL);
+        const isolateCountProp: IsolateCountedDTO = await isolateCountResponse.json();
+
+        getTableContext(isolateCountProp);
+    };
+
```
GerberI: Consider adding an api service that does the fetching and mapping of the data. So that you have only one point in the code where the api fetching stuff happens. Also you don't want handle the DTO conversion stuff in your components but in an encapsulated part of the code. Every time you import some DTO model into a component it is kind of a code smell for me.
So to make it clear: The api service should be the layer that separates the api calls and dto handling from your components and internal data structures.

GerberI (approved):
Commit: Parse counted isolates form api to table format

GerberI (approved):
Commit: Remove unused services & combine PathServices
Good.
If the app grows it could be useful to move the feature dependent functionality to the actual feature that's using it. So that the core services does only contain feature independent functionalities. So the generatePathString.service is actuatlly feature independent, the others are not.

GerberI (approved):
Commit: Set NrOfIsolates inside Content-Results
So i would have preferred the other way around. The container should be responsible for fetching data from the api, not the view component showing the results.

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -126,9 +132,19 @@ export function ContentResultsComponent(): JSX.Element {
     ]);
 
     return (
-        <ResultsTableResultsComponent
-            displayRowCol={state}
-            columnAttributes={columnAttributes}
+        <QueryPageLoadingOrErrorComponent
+            status={{
```
GerberI: I commented in a previous ticket it would be a good idea to refactor the LoadingErrorComponent so that you directly say if it should be active like LoadingErrorComponent isLoading = true.
So now you see why: setting the status of the api endpoints to 200 here is more like a hack than clean code. Also you have to add for each new api endpoint a new status prop.

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -106,13 +108,17 @@ export function ContentResultsComponent(): JSX.Element {
...
-            nrOfSelectedIsolates: isolateCountProp.totalNumberOfIsolates,
-        });
+        if (isolateCountStatus === 200) {
+            const isolateCountProp: IsolateCountedDTO = await isolateCountResponse.json();
+            getTableContext(isolateCountProp);
+            setData({
+                ...data,
+                nrOfSelectedIsolates: isolateCountProp.totalNumberOfIsolates,
+            });
+        }
```
GerberI: The additional status checking and errorPage loading here is another reason why it is better to place the fetching stuff in the container (as stated in a previous commit)

GerberI (changes requested):
Commit: Check api status

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
GerberI: Obsolete

src/app/Shared/LoadingOrError.component.tsx
```diff
@@ -1,14 +1,14 @@
 import React from "react";
-import { ErrorPageComponent } from "../ErrorPage/ErrorPage.component";
-import { LoadingProcessComponent } from "../../../Shared/LoadingProcess.component";
+import { ErrorPageComponent } from "../Main/Body/ErrorPage/ErrorPage.component";
```
GerberI, line 2: Reject: Shared components should never ever import something from your features or main module. So move the ErrorPage to shared or make the component to display on error parameterizable.

GerberI (changes requested):
Commit: Refactoring
Good: Moving the logic up the hierarchy to the container and making the view components more stupid.
Bad: Making a container component a child of a view component: View components should only contain other view components. Container components can contain view or other container components. If you ignore this pattern it will be difficult to maintain the code later. E.g. your view components are not reusable anymore if they contain containers.

src/app/Shared/Context/DataContext.tsx
MicheleK17: Yes it is.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -0,0 +1,88 @@
...
+        setting.stat,
+        ISOLATE_FILTERED_URL,
+    ]);
+
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
+        setSetting({ ...setting, [event.target.name]: event.target.checked });
+    };
+
+    return (
+        <HeaderExportComponent settings={setting} handleChange={handleChange} />
```
GerberI, line 107: Later Refactoring consideration: Let the headerExportComponent be more stupid. It could just return what was checked when the user clicked export. So the container should handle the actually export:
`<HeaderExportComponent settings={settings} onExport={(stat: boolean, raw: boolean) => exportService.export(settings, stat, raw} />`

GerberI (changes requested):
Commit: Fetch filtered isolates in header Container
You putted api stuff to a container: good.
As stated in a previous commit: Now your HeaderComponent should be also a container, it contains the HeaderExportContainer.
As a reminder: there are also layout components, they can contain containers, views and layouts. The components they should display are given as a parameter to them. So your HeaderComponent could be a HeaderLayoutComponent

GerberI:
I think we should discuss the container, layout and view component stuff again, if you like.

src/app/Shared/Context/DataContext.tsx
GerberI: Now that ZNDataFiltered is removed: Is this always the same as ZNData.length? If so, i don't understand what's the use of the nrOfSelectedIsolates

GerberI (approved):
Commit: Remove useless ZNDataFiltered from DataContext

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -31,42 +31,63 @@ export function HeaderExportContainerComponent(): JSX.Element {
...
-        const rawData: DbCollection = raw ? adaptedFilteredIsolates : [];
-        const rawKeys: DbKey[] = raw ? data.keyValues : [];
-        let statData: Record<string, string>[] = [];
-        let statKeys: string[] = [];
-        if (stat) {
-            if (table.option === "absolute") {
-                statData = table.statisticDataAbsolute;
+        const isolateFilteredStatus = isolateFilteredResponse.status;
+
+        if (isolateFilteredStatus === 200) {
```
GerberI, line 37: Can be simplified:
```
rawData, rawKeys = [],
statData, statKeys = []
if(isolateFilteredStatus===200){
  if(raw){
    setRawDataAndKeys
  }
  if(stat) {
    setStatDataAndKeys
  }
}
setSettings(...)
```

GerberI (approved):
Commit: Check status of filtered isolates
So if the fetching fails nothing is exported?
General Consideration:
You fetch data from the api at different points in the app. If you do so it is possible that your whole app state becomes inconsistent. E.g. if you display and filter your data and then export it, it might be possible that the data on the server has changed, so that the exported data is not the same as the data that is displayed in the app. You should consider to fetch the data relevant to a feature (there is only one now) at one point in the app (queryPageContainer). Here it should be decided when new data is fetched (e.g. a user changes a filter setting). All the components should then use this same fetched data. If you want that sub components are allowed to trigger a fetch they should all call the same service, that fetches the data and updates the context for the app.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
MicheleK17: Done. Commit "Make ist simple"

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: Done. Commit "Make ist simple"

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
```diff
@@ -1,126 +1,125 @@
...
+                                (isolateGroup[colAttribute] === colElement ||
+                                    isolateGroup[colAttribute] === undefined)
+                            ) {
+                                const statTableKey =
+                                    isolateGroup[colAttribute] === undefined
+                                        ? t("Results.TableHead")
+                                        : isolateGroup[colAttribute];
+                                tempStatTable[
+                                    statTableKey
+                                ] = (isolateGroup.count as unknown) as string;
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Make ist simple"

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
MicheleK17: Done. Commit "Make it simple"

src/app/Shared/LoadingOrError.component.tsx
MicheleK17: Done. Commit "Make it simple"

src/app/Core/FilterServices/checkIfFilterIsSet.service.ts
```diff
@@ -1,23 +1,24 @@
...
-    FilterInterface,
-    FilterType,
-} from "../../Shared/Model/Filter.model";
+import { FilterInterface, FilterType } from "../../Shared/Model/Filter.model";
 
 /**
  * @desc Check if one of all possible filters is selected
  * @param {FilterInterface} filter - object of filters
+ * @param {string[]} mainFilterAttributes - list with all main filters
  * @returns {boolean} - true if no filter is selected
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Make it simple"

src/app/Core/FilterServices/checkIfFilterIsSet.service.ts
```diff
@@ -1,23 +1,24 @@
...
  */
-export function CheckIfFilterIsSet(filter: FilterInterface, mainFilterAttributes: string[]): boolean {
-    const noFilter = mainFilterAttributes.every(function emptyArray(
+export function CheckIfFilterIsSet(
+    filter: FilterInterface,
+    mainFilterAttributes: string[]
+): boolean {
+    const isFilter = !mainFilterAttributes.every(function emptyArray(
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Make it simple"

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
MicheleK17: Done. Commit "Rename rowsWithPercent"

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -115,6 +123,7 @@ export function QueryPageContainerComponent(): JSX.Element {
         <QueryPageLoadingOrErrorComponent
             status={status}
             dataIsSet={!_.isEmpty(data.ZNData)}
+            componentToDisplay={<QueryPageComponent />}
```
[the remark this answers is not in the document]
MicheleK17: Yes, this is one option to pass a child component in react. And I render the QueryPage only under certain conditions, that's why I chose to pass it as a prop.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: In the final version this function is no longer in the queryPage container. So this comment is obsolete.

src/app/Shared/Context/DataContext.tsx
MicheleK17: Question answered in stand-up.

src/app/Main/Body/QueryPage/QueryPage-LoadingOrError.component.tsx
```diff
@@ -13,24 +12,31 @@ export function QueryPageLoadingOrErrorComponent(props: {
     status:
         | {
               isolateStatus: number;
+              isolateCountStatus: number | undefined;
```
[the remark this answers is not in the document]
MicheleK17: It is because the two current containers handle the data in different ways. This will change with the refactoring of the container-view-structure.

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
MicheleK17: This will change with the refactoring of the container-view-structure.

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
MicheleK17: This will change with the refactoring of the container-view-structure.

src/app/Main/Header/Export/Header-Export.component.tsx
```diff
@@ -19,65 +12,21 @@ import {
 } from "../../../Core/ExportServices/generateExportLabels.service";
 import { ExportButtonLabelComponent } from "./Export-ButtonLabel.component";
 
-export function HeaderExportComponent(): JSX.Element {
+export function HeaderExportComponent(props: {
+    settings: ExportInterface;
+    handleChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Make the header export components more stupid."

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
MicheleK17: (partial?) Done. Commit "Make the header export components more stupid."

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
MicheleK17: This will change with the refactoring of the container-view-structure.

GerberI (approved):
Commit: Make it simple
Thx, looks good

GerberI (approved):
Commit: Rename "rowsWithPercent"
thx

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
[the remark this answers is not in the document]
GerberI: you're right. But as stated before, i would move the checking stuff to the container and decide here what to display. So the loadingorerror component becomes more stupid...

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
MicheleK17: I think it makes no sense to start the iteration over the isolateCountGroups, because in each iteration a row is created for the table. So I need one iteration for each rowValue - that's why I start with that.

src/app/Main/Header/Export/Export-ActionButtons.component.tsx
```diff
@@ -51,11 +51,7 @@ export function ExportActionButtonsComponent(
 ): JSX.Element {
     const { t } = useTranslation(["Export"]);
 
-    const handleClose = (event: unknown): void => {
-        props.onClick(event);
-    };
-
-    const fileIsSelect = !(!props.setting.raw && !props.setting.stat);
```
GerberI, line 58: hehe, boolean logic operations, reminds me of university...

GerberI (changes requested):
Commit: Make the header export components more stupid.
Thx, it looks more like it should look like.
Some suggestions to make the code cleaner:
- I would rename the headerExport component to exportDialog and move the headerExportButton to the container. Because the header button and the dialog are two different things and it makes no sense to encapsulate them in one view.
- I would not split the now headerDialog into exportActionButtons and exportCheckBoxes. The exportDialog should encapsulate both checkboxes and buttons because they stick together.
- The exportDialog should have an onExport event with the state of the checkboxes and the container should then fetch the necessary data and export it.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
GerberI: yes partially done

src/app/Shared/URLs.ts
MicheleK17: Done. Commit: "Change upper case style and remove "event""

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: Done. Commit: "Change upper case style and remove "event""

src/app/Main/Header/Export/Export-ActionButtons.component.tsx
```diff
@@ -24,7 +24,7 @@ const warningStyle = (isSelect: boolean): SerializedStyles => css`
 `;
 
 export interface ExportActionButtonProps {
-    onClick: (event: unknown) => void;
+    handleClickClose: (event: unknown) => void;
```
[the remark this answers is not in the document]
MicheleK17: 1. Part Done. Commit: "Change upper case style and remove "event""

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -30,7 +30,7 @@ export function HeaderExportContainerComponent(): JSX.Element {
     const { filter } = useContext(FilterContext);
     const history = useHistory();
 
-    const ISOLATE_FILTERED_URL: string = isolateURL + history.location.search;
+    const ISOLATE_FILTERED_URL: string = ISOLATE_URL + history.location.search;
```
GerberI, line 33: same here: isolateFilteredUrl

GerberI (approved):
Change upper case style and remove "event"

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
MicheleK17: Done.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -86,11 +94,29 @@ export function HeaderExportContainerComponent(): JSX.Element {
         ISOLATE_FILTERED_URL,
     ]);
 
-    const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
-        setSetting({ ...setting, [event.target.name]: event.target.checked });
+    const handleChange = (name: string, checked: boolean): void => {
+        setSetting({ ...setting, [name]: checked });
```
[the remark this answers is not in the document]
MicheleK17: As we have already agreed, I will take care of these container-issues in an extra ticket. There is no need to discuss it further in this CR.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
MicheleK17: Done.

src/app/Main/Header/Export/Header-Export.component.tsx
```diff
@@ -1,60 +1,50 @@
...
     const { filter } = useContext(FilterContext);
 
-    const handleClickOpen = (): void => {
-        setOpen(true);
-    };
-    const handleClose = (): void => {
-        setOpen(false);
-    };
-
-    const buttonLabel: JSX.Element = ExportButtonLabelComponent(open);
```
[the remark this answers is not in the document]
MicheleK17: isOpen makes a difference in the sytle of the ExportButtonLabel.

src/app/Main/Header/Export/Export-ActionButtons.component.tsx
```diff
@@ -68,11 +64,11 @@ export function ExportActionButtonsComponent(
         <div>
             <p css={warningStyle(fileIsSelect)}>{t("Warning")}</p>
             <DialogActions>
-                <Button onClick={handleClose} color="primary">
+                <Button onClick={props.handleClickClose} color="primary">
                     {t("Button.Cancel")}
                 </Button>
                 <Button
-                    onClick={handleClose}
+                    onClick={props.handleClickClose}
```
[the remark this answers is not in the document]
MicheleK17: As we have already agreed, I will take care of these container-issues in an extra ticket. There is no need to discuss it further in this CR.

src/app/Main/Header/Export/Export-ActionButtons.component.tsx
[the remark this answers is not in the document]
MicheleK17: 2. Part: I don't think "cancel" is ab better name, because the dialog is also closed after choosing "export".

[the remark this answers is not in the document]
MicheleK17: 3. Part: Done. Commit: "Change handleClick"

GerberI (approved):
Commit: Change ISOLATE_FILTERED_URL to isolateFilteredUrl
thx

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -98,10 +98,10 @@ export function HeaderExportContainerComponent(): JSX.Element {
         setSetting({ ...setting, [name]: checked });
     };
 
-    const handleClickOpen = (): void => {
+    const onClickOpen = (): void => {
```
GerberI, line 101: one rename too much: this was correct: handleClickOpen
same below

src/app/Main/Header/Export/Header-Export.component.tsx
```diff
@@ -13,21 +13,21 @@ export function HeaderExportComponent(props: {
...
+    onClickClose: () => void;
     handleCheckbox: (name: string, checked: boolean) => void;
 }): JSX.Element {
     const { filter } = useContext(FilterContext);
 
     return (
         <div>
             <HeaderExportButtonComponent
-                handleClickOpen={props.handleClickOpen}
+                onClickOpen={props.onClickOpen}
```
GerberI, line 25: you could use handleClick function in this file to. So you have the same convention in all files

GerberI (changes requested):
Commit: Change handleClick

src/app/Main/Header/Export/Export-ActionButtons.component.tsx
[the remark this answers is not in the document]
GerberI: thx, for the 2.Part: yes, you're right, i was thinking of having a separate onExportClick event and the container handling the closing. Then the other button would be onClickCancel

src/app/Main/Header/Export/Export-ActionButtons.component.tsx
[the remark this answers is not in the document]
GerberI: i know, it was only a suggestion how it can look like later.

src/app/Main/Header/Export/Header-Export.component.tsx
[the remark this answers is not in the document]
GerberI: i see isOpen is used to disable hover color if opened. But you cannot hover the button if the dialog is opened, so i don't understand why it is used.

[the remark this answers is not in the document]
MicheleK17: The button is used in two different places. Once in the header (here hoover is possible) and once in the dialog (dialog open = here no hoover possible).

src/app/Main/Body/QueryPage/Results/Content-Results.component.tsx
GerberI: Discussed in meeting

GerberI (approved):
Commit: Change handleClick, move ExportLabels
thx

GerberI:
All done, ready to merge 🥇

PR #227, opened 2021-03-09
Add all main Filters (total: 10)

MicheleK17:
String-Filter: 
    - federalState
    - samplingStage
    - origin
    - category
    - productionType

Date-Filter: 
    - sampling year
   
Object-Filter:
    - antimicrobial resistance

### DIFF
--- a/src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
+++ b/src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
@@ -1,0 +1,29 @@
+import { MainFilterLabelInterface } from "../../../Shared/Model/Export.model";
+import {
+    DbCollection,
+    DbKey,
+} from "../../../Shared/Model/Client_Isolate.model";
+import { generateDataCsvString } from "./generateDataCsvString.service";
+
+/**
+ * @desc Returns the table header and the filtered dataset as string to save it as csv
+ * @param {string[]} keys - keys of the dataset
+ * @param {Record<string, string>[]} data - dataset
+ * @returns {string} - dataset as string
+ */
+export function generateDataString(
+    data: DbCollection,
+    keys: DbKey[], 
+    mainFilterLabels: MainFilterLabelInterface,
+    mainFilterAttributes: string[],
+): string {
+    const FilteredDataString: string[] = [];
+    const headerArray: string[] = []
+    mainFilterAttributes.forEach(element => {
+        headerArray.push(mainFilterLabels[element])
+    });
+    FilteredDataString.push(headerArray.join(","));
+    FilteredDataString.push(generateDataCsvString(data, keys));
+
+    return FilteredDataString.join("\n");
+}
--- a/src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
+++ b/src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
@@ -9,18 +9,20 @@ import { generateDataCsvString } from "./generateDataCsvString.service";
  * @desc Returns the table header and the filtered dataset as string to save it as csv
  * @param {string[]} keys - keys of the dataset
  * @param {Record<string, string>[]} data - dataset
+ * @param {MainFilterLabelInterface} mainFilterLabels - Obj with translated labels for the table header
+ * @param {string[]} mainFilterAttributes - keys to get matching mainFilterLabel
  * @returns {string} - dataset as string
  */
 export function generateDataString(
     data: DbCollection,
-    keys: DbKey[], 
+    keys: DbKey[],
     mainFilterLabels: MainFilterLabelInterface,
-    mainFilterAttributes: string[],
+    mainFilterAttributes: string[]
 ): string {
     const FilteredDataString: string[] = [];
-    const headerArray: string[] = []
-    mainFilterAttributes.forEach(element => {
-        headerArray.push(mainFilterLabels[element])
+    const headerArray: string[] = [];
+    mainFilterAttributes.forEach((element) => {
+        headerArray.push(mainFilterLabels[element]);
     });
     FilteredDataString.push(headerArray.join(","));
     FilteredDataString.push(generateDataCsvString(data, keys));
--- a/src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
+++ b/src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
@@ -7,8 +7,8 @@ import { generateDataCsvString } from "./generateDataCsvString.service";
 
 /**
  * @desc Returns the table header and the filtered dataset as string to save it as csv
- * @param {string[]} keys - keys of the dataset
- * @param {Record<string, string>[]} data - dataset
+ * @param {DbCollection} data - dataset
+ * @param {DbKey[]} keys - keys of the dataset
  * @param {MainFilterLabelInterface} mainFilterLabels - Obj with translated labels for the table header
  * @param {string[]} mainFilterAttributes - keys to get matching mainFilterLabel
  * @returns {string} - dataset as string
@@ -20,10 +20,9 @@ export function generateDataString(
     mainFilterAttributes: string[]
 ): string {
     const FilteredDataString: string[] = [];
-    const headerArray: string[] = [];
-    mainFilterAttributes.forEach((element) => {
-        headerArray.push(mainFilterLabels[element]);
-    });
+    const headerArray: string[] = mainFilterAttributes.map(
+        (element) => mainFilterLabels[element]
+    );
     FilteredDataString.push(headerArray.join(","));
     FilteredDataString.push(generateDataCsvString(data, keys));
 
--- a/src/app/Core/ExportServices/StatExportServices/generateStatisticString.service.ts
+++ b/src/app/Core/ExportServices/StatExportServices/generateStatisticString.service.ts
@@ -1,5 +1,5 @@
 import _ from "lodash";
-import { ExportInterface } from "../../../Shared/Model/Export.model";
+import { FilterType } from "../../../Shared/Model/Filter.model";
 import { generateStatisticCsvString } from "./generateStatisticCsvString.service";
 
 /**
@@ -9,23 +9,24 @@ import { generateStatisticCsvString } from "./generateStatisticCsvString.service
  * @returns {string} - header and statistic table as string
  */
 export function generateStatisticString(
-    exportParameter: ExportInterface
+    tableAttributes: { row: FilterType; column: FilterType },
+    statDataSet: { statData: Record<string, string>[]; statKeys: string[] }
 ): string {
     const StatDataString: string[] = [];
-    if (!_.isEmpty(exportParameter.tableAttributes.column)) {
-        StatDataString.push(`,${exportParameter.tableAttributes.column}`);
+    if (!_.isEmpty(tableAttributes.column)) {
+        StatDataString.push(`,${tableAttributes.column}`);
     }
-    const headers: string[] = exportParameter.statDataSet.statKeys;
-    if (!_.isEmpty(exportParameter.tableAttributes.row)) {
+    const headers: string[] = statDataSet.statKeys;
+    if (!_.isEmpty(tableAttributes.row)) {
         const headerToPrint: string[] = [...headers];
-        headerToPrint[0] = exportParameter.tableAttributes.row;
+        headerToPrint[0] = tableAttributes.row;
         StatDataString.push(headerToPrint.join(","));
     } else {
         StatDataString.push(headers.join(","));
     }
 
     StatDataString.push(
-        generateStatisticCsvString(exportParameter.statDataSet.statData, headers)
+        generateStatisticCsvString(statDataSet.statData, headers)
     );
 
     return StatDataString.join("\n");
--- a/src/app/Core/ExportServices/StatExportServices/generateStatisticString.service.ts
+++ b/src/app/Core/ExportServices/StatExportServices/generateStatisticString.service.ts
@@ -1,6 +1,6 @@
 import _ from "lodash";
-import { ExportInterface } from "../../Shared/Model/Export.model";
-import { generateCSVString } from "./generateCSVString.service";
+import { ExportInterface } from "../../../Shared/Model/Export.model";
+import { generateStatisticCsvString } from "./generateStatisticCsvString.service";
 
 /**
  * @desc Returns the table header and the statistic table as a string to save it as CSV
@@ -8,7 +8,7 @@ import { generateCSVString } from "./generateCSVString.service";
  * @param {{statData: Record<string, string>[]; statKeys: string[];}} statDataSet - statistic table
  * @returns {string} - header and statistic table as string
  */
-export function generateStatDataString(
+export function generateStatisticString(
     exportParameter: ExportInterface
 ): string {
     const StatDataString: string[] = [];
@@ -25,7 +25,7 @@ export function generateStatDataString(
     }
 
     StatDataString.push(
-        generateCSVString(exportParameter.statDataSet.statData, headers)
+        generateStatisticCsvString(exportParameter.statDataSet.statData, headers)
     );
 
     return StatDataString.join("\n");
--- a/src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
+++ b/src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
@@ -4,9 +4,9 @@ import {
     MainFilterLabelInterface,
 } from "../../Shared/Model/Export.model";
 import { generateParameterHeader } from "./generateParameterHeader.service";
-import { generateStatDataString } from "./generateStatDataString.service";
+import { generateStatisticString } from "./StatExportServices/generateStatisticString.service";
 import { DbKeyCollection } from "../../Shared/Model/Client_Isolate.model";
-import { generateDataString } from "./generateDataString.service";
+import { generateDataString } from "./DataExportServices/generateDataString.service";
 
 export interface ObjectToCsvParameter {
     setting: ExportInterface;
@@ -24,7 +24,7 @@ export interface ObjectToCsvParameter {
  * @param {MainFilterLabelInterface} mainFilterLabels - object with labels of the main filters
  * @returns {string} - converted data as csv string
  */
-export function parameterAndDataToCsvString(csvParameter: ObjectToCsvParameter): string {
+export function dataOrStatisticToCsvString(csvParameter: ObjectToCsvParameter): string {
     const csvRows: string[] = [];
     csvRows.push(
         generateParameterHeader(
@@ -47,7 +47,7 @@ export function parameterAndDataToCsvString(csvParameter: ObjectToCsvParameter):
     }
 
     if (csvParameter.setting.stat && !csvParameter.setting.raw) {
-        csvRows.push(generateStatDataString(csvParameter.setting));
+        csvRows.push(generateStatisticString(csvParameter.setting));
     }
 
     return csvRows.join("\n");
--- a/src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
+++ b/src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
@@ -18,11 +18,7 @@ export interface ObjectToCsvParameter {
 
 /**
  * @desc Convert the chosen parameter and the resulting Data to a CSV string
- * @param {ExportInterface} setting - all info for export (raw/stat, row&column, dataset)
- * @param {FilterInterface} filter - object with the selected filters
- * @param {string} allFilterLabel - "all values" / "Alle Werte"
- * @param {MainFilterLabelInterface} mainFilterLabels - Obj with translated labels for the table header
- * @param {string[]} - keys to get matching mainFilterLabel
+ * @param {ObjectToCsvParameter} csvParameter
  * @returns {string} - converted data as csv string
  */
 export function dataOrStatisticToCsvString(
--- a/src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
+++ b/src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
@@ -4,9 +4,9 @@ import {
     MainFilterLabelInterface,
 } from "../../Shared/Model/Export.model";
 import { generateParameterHeader } from "./generateParameterHeader.service";
-import { generateStatisticString } from "./StatExportServices/generateStatisticString.service";
+import { generateStatisticTableCsvString } from "./StatExportServices/generateStatisticTableCsvString.service";
 import { DbKeyCollection } from "../../Shared/Model/Client_Isolate.model";
-import { generateDataString } from "./DataExportServices/generateDataString.service";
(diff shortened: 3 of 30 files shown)


src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -2,8 +2,37 @@ export type DbCollection = {
     microorganism: string;
     samplingContext: string;
     matrix: string;
+    federalState: string;
+    samplingStage: string;
+    origin: string;
+    category: string;
+    productionType: string;
```
GerberI: could be: `export type DBCollection = Record<DbKey, string>;`

src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -2,8 +2,37 @@ export type DbCollection = {
...
+    "samplingContext",
+    "matrix",
+    "federalState",
+    "samplingStage",
+    "origin",
+    "category",
+    "productionType",
+];
+
+export interface ClientIsolateCountedDTO {
```
GerberI: Why is this here? It's a DTO and should be where the API models are.

GerberI (changes requested):
Commit: Add simple string-filter

src/app/Shared/Model/Client_Isolate.model.ts
MicheleK17: It's done in the commit "Add sampling year filter".

src/app/Shared/Model/Client_Isolate.model.ts
MicheleK17: It's done in the commit "Implement resistance-filter ".

src/app/Core/ExportServices/generateDataString.service.ts
```diff
@@ -7,13 +46,18 @@ import { generateCSVString } from "./generateCSVString.service";
...
+    mainFilterAttributes: string[],
 ): string {
     const FilteredDataString: string[] = [];
-    const headers: string[] = keys;
-    FilteredDataString.push(headers.join(","));
-    FilteredDataString.push(generateCSVString(data, headers));
+    const headerArray: string[] = []
+    mainFilterAttributes.forEach(element => {
+        headerArray.push(mainFilterLabels[element])
+    });
```
GerberI: could be simplified: headerArray = mainFilterAttributes.map(element =>mainFiltrrLabels[element])

src/app/Core/ExportServices/modifyTableDataString.service.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @desc escape all quotes and replaces "undefined" with an empty string
+ * @param {string} inputString - string to be modified
+ * @returns {string} - modified string
+ */
+export function modifyTableDataStringService(inputString: string): string {
+    return `"${inputString.replace(/"/g, '\\"').replace("undefined", "")}"`;
```
GerberI, line 7: why not return inputString.replace(...).replace(...)

src/app/Core/adaptIsolatesFromAPI.service.ts
```diff
@@ -21,6 +22,7 @@ export function adaptIsolatesFromAPI(isolateProp: IsolateDTO): DbCollection {
             origin,
             category,
             productionType,
+            resistance: Object.keys(resistance) as ResistantValues[]
```
GerberI: Reminder: the type conversion here will not prevent the running code to add faulty values to resistance because the DTO data comes from an external source and cannot be checked during compile time. The code will fail elsewhere when you try to access the resistance object. So a more robust way would be:
```
possibleResistancesArray.foreach(res => { 
  if(dto.resistance.res !== undefined) {
    model.resistance.push(res)
  }
})
```

src/app/Shared/Model/Api_Isolate.model.ts
```diff
@@ -23,7 +23,7 @@ export interface IsolateDTO {
             e_hly: string;
             ampc_carba_phenotype: string;
         };
-        resistences: {
+        resistance: {
```
GerberI, line 26: why is this singular now? Perhaps you should use resistances in your client model

src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -1,13 +1,53 @@
...
-    microorganism: string;
-    samplingContext: string;
-    matrix: string;
-    federalState: string;
-    samplingStage: string;
-    origin: string;
-    category: string;
-    productionType: string;
-}[];
+export type ResistantValues =
```
GerberI, line 1: perhaps ResistanceKey?

src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -1,13 +1,53 @@
...
+    | "tec"
+    | "tet"
+    | "tgc"
+    | "tia"
+    | "tmp"
+    | "trm"
+    | "t_c"
+    | "van";
+
+export type DbStringKey =
```
GerberI, line 38: whats a DbStringKey? Perhaps you mean something like DbPrimaryAttributeKey or something?
Can be simlified: `= Exclude<DbKey, "resistance">`

GerberI (approved):
Commit: Add all main Filters (total: 10)
Good, only minor issues

src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
```diff
@@ -0,0 +1,29 @@
...
+} from "../../../Shared/Model/Client_Isolate.model";
+import { generateDataCsvString } from "./generateDataCsvString.service";
+
+/**
+ * @desc Returns the table header and the filtered dataset as string to save it as csv
+ * @param {string[]} keys - keys of the dataset
+ * @param {Record<string, string>[]} data - dataset
+ * @returns {string} - dataset as string
+ */
+export function generateDataString(
```
GerberI, line 16: this also generates a csv string. Perhaps this could be generateTableCsvString

src/app/Core/ExportServices/StatExportServices/generateStatisticString.service.ts
```diff
@@ -1,14 +1,14 @@
...
+import { generateStatisticCsvString } from "./generateStatisticCsvString.service";
 
 /**
  * @desc Returns the table header and the statistic table as a string to save it as CSV
  * @param {{row: FilterType; column: FilterType;}} tableAttributes - selected row and column
  * @param {{statData: Record<string, string>[]; statKeys: string[];}} statDataSet - statistic table
  * @returns {string} - header and statistic table as string
  */
-export function generateStatDataString(
+export function generateStatisticString(
```
GerberI, line 11: same here: perhaps generateStatisitcTableCsvString

GerberI (approved):
Commit: Little refactoring ExportServices
General consideration: In general services should be encapsulated into singleton classes. That is the definition of a service. What you have here are loose functions like in the old days where object oriented way was not invented yet. The idea is that you encapsulate all the stuff in one class. One advantage  is that you do not to need to invent fancy names for your functions:
```
class CsvService {
  public convertStatisticsToCsv(...)
  public convertTableToCsv(...)
  your helper functions becomes private and are not visible to the outside anymore:
  private convertStatisitcDataToCsv(...)
  ...
}
class ZipService {
  public generateZipFromCsv(...) {
    ...
    csvData = CsvService.convertTableToCsv(...)
    ...
  }
}
```

src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
```diff
@@ -9,18 +9,20 @@ import { generateDataCsvString } from "./generateDataCsvString.service";
  * @desc Returns the table header and the filtered dataset as string to save it as csv
  * @param {string[]} keys - keys of the dataset
  * @param {Record<string, string>[]} data - dataset
+ * @param {MainFilterLabelInterface} mainFilterLabels - Obj with translated labels for the table header
```
GerberI, line 12: and DbKey[], DbCollection for key and data

src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
```diff
@@ -21,10 +21,13 @@ export interface ObjectToCsvParameter {
  * @param {ExportInterface} setting - all info for export (raw/stat, row&column, dataset)
  * @param {FilterInterface} filter - object with the selected filters
  * @param {string} allFilterLabel - "all values" / "Alle Werte"
- * @param {MainFilterLabelInterface} mainFilterLabels - object with labels of the main filters
+ * @param {MainFilterLabelInterface} mainFilterLabels - Obj with translated labels for the table header
+ * @param {string[]} - keys to get matching mainFilterLabel
```
GerberI, line 25: this is the description for the ObjectToCsvParameter interface, the function only has one "csvParameter"

GerberI (approved):
Commit: Adapt function desc. in the ExportService

src/app/Core/ExportServices/generateDataString.service.ts
```diff
@@ -1,4 +1,43 @@
...
+    DbKey,
+    DbStringKey,
+    ResistantValues,
+} from "../../Shared/Model/Client_Isolate.model";
+import { modifyTableDataStringService } from "./modifyTableDataString.service";
+
+/**
+ * @desc Transform the data to a string for the CSV file. Elements of one row are separated by comma.
+ * @param {Record<string, string>[]} dataArray - data object to export in a csv file
+ * @param {k[]} headers - keys of the data
```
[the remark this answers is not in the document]
GerberI: done

src/app/Core/ExportServices/generateDataString.service.ts
```diff
@@ -1,4 +1,43 @@
...
+ * @returns {string} - data in one csv string
+ */
+function generateDataCSVString(
+    dataArray: DbCollection,
+    headers: DbKey[]
+): string {
+    const csvTable: string[] = [];
+    dataArray.forEach(
+        (
+            row: Record<DbStringKey, string> & { resistance: ResistantValues[] }
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Simplify code"

src/app/Core/ExportServices/generateDataString.service.ts
```diff
@@ -1,4 +1,43 @@
...
+        ) => {
+            const oneRow: string[] = [];
+            headers.forEach((element) => {
+                if (element !== "resistance") {
+                    const rowValue = String(row[element]);
```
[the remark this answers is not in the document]
MicheleK17: I have no idea why I did this here. Maybe I put the translation here earlier. 
Done. Commit "Simplify code"

src/app/Core/ExportServices/generateDataString.service.ts
```diff
@@ -1,4 +1,43 @@
...
+function generateDataCSVString(
+    dataArray: DbCollection,
+    headers: DbKey[]
+): string {
+    const csvTable: string[] = [];
+    dataArray.forEach(
+        (
+            row: Record<DbStringKey, string> & { resistance: ResistantValues[] }
+        ) => {
+            const oneRow: string[] = [];
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Simplify code"

src/app/Core/ExportServices/generateDataString.service.ts
```diff
@@ -1,4 +1,43 @@
...
+            headers.forEach((element) => {
+                if (element !== "resistance") {
+                    const rowValue = String(row[element]);
+                    oneRow.push(modifyTableDataStringService(`${rowValue}`));
```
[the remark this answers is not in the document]
MicheleK17: I have no idea why I did this here. Maybe I put the translation here earlier.
Done. Commit "Simplify code"

src/app/Core/ExportServices/generateDataString.service.ts
MicheleK17: Done. Commit "Simplify code"

src/app/Shared/Model/Api_Isolate.model.ts
MicheleK17: Done. Commit "Simplify code"

src/app/Shared/Model/Client_Isolate.model.ts
MicheleK17: These values aren't used as keys. That's why I don't think renaming is useful.

src/app/Shared/Model/Client_Isolate.model.ts
MicheleK17: Done. Commit "Simplify code"

src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
MicheleK17: Done.

src/app/Core/ExportServices/dataOrStatisticToCsvString.service.ts
MicheleK17: Done. Commit "Simplify code"

src/app/Core/ExportServices/generateDataString.service.ts
```diff
@@ -1,4 +1,43 @@
...
+} from "../../Shared/Model/Client_Isolate.model";
+import { modifyTableDataStringService } from "./modifyTableDataString.service";
+
+/**
+ * @desc Transform the data to a string for the CSV file. Elements of one row are separated by comma.
+ * @param {Record<string, string>[]} dataArray - data object to export in a csv file
+ * @param {k[]} headers - keys of the data
+ * @returns {string} - data in one csv string
+ */
+function generateDataCSVString(
```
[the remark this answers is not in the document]
MicheleK17: Will be taken into account when refactoring the container-view-structure.

src/app/Core/ExportServices/modifyTableDataString.service.ts
MicheleK17: It is necessary to include commas within the filter values and not splitting them in the csv file.

src/app/Core/ExportServices/DataExportServices/generateDataString.service.ts
MicheleK17: Done. Commit: Rename "generate...CsvString.service"

src/app/Core/ExportServices/StatExportServices/generateStatisticString.service.ts
MicheleK17: Done. Commit: Rename "generate...CsvString.service"

src/app/Shared/Model/Api_Isolate.model.ts
MicheleK17: For this change I would need to adapt every data I get from the APIs (filter, isolate, counted isoalte). I can mention it when I talk to Dominic about the API and we can change it in the BE to "resistances".

src/app/Core/adaptIsolatesFromAPI.service.ts
MicheleK17: Based on previous CR, there is already a tick for data validation. However, this does not have a high priority at the moment.

GerberI (changes requested):
Commit: Add sampling year filter
I wonder if the api does not return only strings? And perhaps you could tell your json parser that it should parse only strings and do the conversion stuff for you.
And i wonder why a year get's parsed to a Date object.

GerberI (approved):
Commit: Simplify code
thx

src/app/Shared/Model/Client_Isolate.model.ts
GerberI: then perhaps ResistantCode? ResistantValues sounds like the values are somehow sticky or permanent. Nevertheless the name should be singular because the type stays for only one of the defined strings.

src/app/Core/ExportServices/modifyTableDataString.service.ts
GerberI: ok, i see, so the output is additionally surrounded by quotes? perhaps you should add this to the description of the function.

src/app/Shared/Model/Api_Isolate.model.ts
GerberI: ok, thought is was just a one-liner in a dto mapping function.

GerberI (approved):
Commit: Add all main Filters
cool, thx. The hardest part of the review was to tell and remember what these functions actually do. Now you can see it at a first glance.

src/app/Core/adaptFilterFromAPI.service.ts
```diff
@@ -14,7 +14,7 @@ export function adaptFilterFromApiService(
             id,
             name,
             parent,
-            values: values.join().split(','),
+            values: (values as Array<string|Date>).map((element: string | Date) => String(element)),
```
GerberI, line 17: i thought of adding a comment to use map instead, now you've done it yourself :)
no need to type it here, i think you typed it already in FilterConfigDTO and ClientFilterConfig. So just `values: values.map(value=>String(value))` would do it.

GerberI (approved):
Commit: Debug Filter-Values with comma

src/app/Core/adaptFilterFromAPI.service.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function adaptFilterFromApiService(
+    filterProp: FilterConfigDTO
+): ClientFilterConfig {
+    const adaptedFilters: ClientFilterConfig = {filters: filterProp.filters.map(
+        ({ id, name, parent, values }) => ({
+            id,
+            name,
+            parent,
+            values: values.join().split(','),
```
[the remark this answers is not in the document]
GerberI: obsolete

src/app/Core/adaptCountedIsolatesGroups.service.ts
```diff
@@ -0,0 +1,36 @@
...
+            const adaptedGroup = _.cloneDeep(element);
+            adaptedGroup.samplingYear = String(adaptedGroup.samplingYear);
+            adaptedGroups.push(
+                adaptedGroup as Record<string, string> & {
+                    count: number;
+                }
+            );
+        } else {
+            adaptedGroups.push(
+                element as Record<string, string> & {
```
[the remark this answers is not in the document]
MicheleK17: Why should I clone it, if I don't change it?

src/app/Core/adaptFilterFromAPI.service.ts
MicheleK17: If I don't type the values as Array I get the following Error: 

This expression is not callable.
  Each member of the union type '(<U>(callbackfn: (value: string, index: number, array: string[]) => U, thisArg?: any) => U[]) | (<U>(callbackfn: (value: Date, index: number, array: Date[]) => U, thisArg?: any) => U[])' has signatures, but none of those signatures are compatible with each other.

src/app/Shared/Model/Filter.model.ts
```diff
@@ -2,7 +2,7 @@ export type FilterType = string;
 
 export type FilterInterface = Record<FilterType, string[]>;
 
-export interface FilterConfigDTO {
+export interface ClientFilterConfig {
```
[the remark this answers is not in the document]
MicheleK17: Done. "Rename & Simplify"

src/app/Core/adaptCountedIsolatesGroups.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    countedIsolatesGroups: (Record<string, string | Date> & {
+        count: number;
+    })[]
+): (Record<string, string> & { count: number })[] {
+    const adaptedGroups: (Record<string, string> & {
+        count: number;
+    })[] = [];
+
+    countedIsolatesGroups.forEach((element) => {
+        if (_.includes(Object.keys(element), "samplingYear")) {
```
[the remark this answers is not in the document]
MicheleK17: thx!

src/app/Core/ExportServices/modifyTableDataString.service.ts
MicheleK17: Done. "Rename & Simplify"

src/app/Shared/Model/Client_Isolate.model.ts
MicheleK17: Done. "Rename & Simplify"

GerberI (approved):
Commit: Rename & Simplify
thx

src/app/Shared/Model/Filter.model.ts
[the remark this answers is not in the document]
GerberI: sorry, the comment was not detailed enough: I meant to use the already defined type ClientSingleFilterConfig to define the type ClientFiltersConfig:
```
interface ClientFiltersConfig {
  filters: ClientSingleFilterConfig[];
}
```

src/app/Core/adaptCountedIsolatesGroups.service.ts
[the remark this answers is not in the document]
GerberI: But you clone it in the if(true) case. So your function clones the data in one case, in the other not. It is inconistent. In general it is a good idea to always clone the data from the API (that would be implicitly the case when you explicitly assign the api data as in the "api data validation ticket"). The reason is, if you have some deep nested data from the api and assign it to clientDataA and clientDataB, and then manipulate clientDataA, clientDataB will automatically changed to. That is often a cause for difficult to trace bugs.

src/app/Shared/Model/Filter.model.ts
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Clone, simplify, debug"

src/app/Core/adaptCountedIsolatesGroups.service.ts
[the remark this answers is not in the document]
MicheleK17: Ah I get it!
Done. Commit "Clone, simplify, debug"

src/app/Core/adaptFilterFromAPI.service.ts
GerberI: I see, typescript generates two mapping function types, so this will not work. But you should be able to simplify to:
```values.map((element: string | Data) => ...)``

Comment (MicheleK17) [reply src/app/Core/adaptFilterFromAPI.service.ts]: No, I have already tried that. I need to type values as an Array.

**Review by GerberI — APPROVED** [review]
Commit: Clone, simplify, debug
thx,
you should name your "debug" ticket comments "fix". Actually you do the debugging and the commit is a fix of the bug :)

File: src/app/Core/adaptFilterFromAPI.service.ts (hunk 19 above)
Comment (GerberI) [reply src/app/Core/adaptFilterFromAPI.service.ts]: Ok, that's strange. Then leave it as it is. In general you would do this also during api data validation and mapping:
````
if(typeof apiRawData === 'Date') {
  myData = parseFromDate(apiRawData)
}
```
Then you do not have to type the api objects before hand, because all you really get from the api is of type unknown and you only tell the compiler how you expect it to look like.
But as mentioned i wonder if you not could tell your json parser to do the conversions for you.

GerberI:
All done, ready to merge.
I try to stick to the convention of only making one non-reject-comment in the next reviews, to make things faster :)

PR #232, opened 2021-03-19
Change Container-View-Structure

MicheleK17:
Ticket: #224
Change-Id: I3142338b650a726738dd0d4cc8394176f0293121


src/app/Shared/Selector.component.tsx
```diff
@@ -79,8 +79,8 @@ export interface SelectorProps {
     selectedValuesObj: { value: string; label: string }[];
     noOptionLabel: string;
     selectAttribute: FilterType | TableType;
-    handleChange: (
-        selectedOption: ValueType<Record<string, string>>,
+    onChange: (
+        selectedOption: {value: string, label: string} | ValueType<{value: string, label: string}>,
```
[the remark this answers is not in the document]
GerberI: General consideration: try to omit using framework or library specific types in your props. The interface of a component should be as simple as possible. E.g. the ValueType is from the library react-select and is rather complex. So you push the complexity to other components who are using this component. So depending on the behaviour of this component (single select vs multi select vs none select allowed, i don't now) the type could be: 
- single select, empty select allowed: `selectedOptions: {value: string, label: string} | null`
- multi select: `selectedOptions: {value: string, label: string}[]`

You would do the type conversion in the local handleChange function:
```
const handleChange = (options: ValueType<...>) => {
  ...some checking of the runtime type of options if necessary...
  props.onDisplFeaturesChange(options as {value: string, label: string}, ...)
}
```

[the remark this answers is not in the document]
GerberI: Following this consideration the simplifying would go on:
E.g. the DisplayedFeatureSelectorComponent would handle the change of the SelectorComponent and convert the data to a more appropriate form, so the mapping that happens in the handleChangeDisplayedFeatures service should actually happen in the DisplayedFeatureSelectorComponent and it should provide an output event like:
onFeatureChange: (name: SomeType, value: SomeType) => void
This goes on through the component tree. Each component provides the data relevant in its context.
So a good practice is too look at each component independently. What would a user expect from this component without knowing the internal components used down the tree. The interface/props of the component should reflect this expectation.

GerberI (changes requested):
Commit: Remove setting Context inside view components
The container now looks more like it should look like (except the api code not extracted to a service). Only by looking through the container code you can understand what the whole feature represented by the container actually does.
The next step could be to remove the table context completely from the views.

src/app/Main/Body/QueryPage/ChangeDisplFeatures.service.tsx
```diff
@@ -0,0 +1,12 @@
+import { ValueType } from "react-select";
+import { FilterType } from "../../../Shared/Model/Filter.model";
+
+export function handleChangeDisplayedFeatures(
+    selectedOption: ValueType<{value: string, label: string}>
+): FilterType {
+    if (selectedOption) {
+        const selectedFeature: string = (Object.values(selectedOption)[0]);
```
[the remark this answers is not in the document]
MicheleK17: I already tried different methods to validate the data in this case. 
Your idea doesn't work because "ValueType" is a type of the selector. Which works equally for the "displayed features" ({ value: string; label: string; }) and for the "filters" (OptionsType<{ value: string; label: string; }>). 

**Property 'value' does not exist on type '{ value: string; label: string; } | OptionsType<{ value: string; label: string; }>'.
  Property 'value' does not exist on type 'OptionsType<{ value: string; label: string; }>'.**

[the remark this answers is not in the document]
MicheleK17: I could type the selectedOption like you suggested in the following comment for the ChangeFilter.service.ts. 

`const selectedOptionObj = selectedOption as {value: string, label: string}`

MicheleK17:
> The next step could be to remove the table context completely from the views.

If I remove the table context completely from the views, I would no longer need a context. Then I would only use props.

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/Drawer-DisplFeatures.component.tsx
```diff
@@ -46,60 +46,40 @@ const iconStyle = css`
...
-export function DrawerDisplayedFeaturesComponent(): JSX.Element {
-    const { table, setTable } = useContext(TableContext);
+interface DrawerDisplayedFeaturesInterface {
+    handleChangeDisplFeatures: (
+        selectedOption: ValueType<{value: string, label: string}>,
+        keyName: FilterType | TableType
+    ) => void;
+    handleSwapDisplFeatures: () => void;
+    handleRemoveAllDisplFeatures: () => void;
+}
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Rename handleChange-Functions, update TypeDocs"

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/Drawer-DisplFeatures.component.tsx
```diff
@@ -46,60 +46,40 @@ const iconStyle = css`
     height: 36px;
 `;
 
-export function DrawerDisplayedFeaturesComponent(): JSX.Element {
-    const { table, setTable } = useContext(TableContext);
+interface DrawerDisplayedFeaturesInterface {
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit: "Rename Interface => Prop"

src/app/Shared/Selector.component.tsx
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Adapt types of selector onChange"

src/app/Main/Body/QueryPage/ChangeFilter.service.ts
```diff
@@ -0,0 +1,14 @@
+import { ValueType } from "react-select";
+
+export function changeFilter(
+    selectedOption: ValueType<{value: string, label: string}>
+): string[] {
+    if (Array.isArray(selectedOption)) {
+        const selectedFilter: string[] = [];
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Adapt types of selector onChange" and "Rename services choose selected values"

src/app/Main/Body/QueryPage/ChangeDisplFeatures.service.tsx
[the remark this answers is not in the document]
MicheleK17: Done. Commit: "Adapt types of selector onChange"

GerberI:
> > The next step could be to remove the table context completely from the views.
> 
> If I remove the table context completely from the views, I would no longer need a context. Then I would only use props.

exactly, see my comment on slack. So i would keep the context, even if it is only used by the container providing it.

src/app/Shared/LoadingOrError.component.tsx
```diff
@@ -11,16 +11,15 @@ import { LoadingProcessComponent } from "./LoadingProcess.component";
...
+              isolateStatus: number | undefined;
               isolateCountStatus: number | undefined;
-              filterStatus: number;
+              filterStatus: number | undefined;
           }
-        | undefined;
     dataIsSet: boolean;
     componentToDisplay: JSX.Element;
 }): JSX.Element {
     if (props.status !== undefined) {
```
GerberI, line 22: reject: according to the props this cannot be undefined anymore

src/app/Shared/LoadingOrError.component.tsx
```diff
@@ -11,16 +11,15 @@ import { LoadingProcessComponent } from "./LoadingProcess.component";
...
-        if (props.status.isolateStatus !== 200) {
+        if (props.status.isolateStatus !== 200 && props.status.isolateStatus) {
```
GerberI, line 23: reject (also below): check explicitly against undefined

GerberI (changes requested):
Commit: Combine QueryPage-Containers, write services
Good, thx

src/app/Main/Body/QueryPage/NumberOfIsolates/QueryPage-NrOfIsolates.component.tsx
```diff
@@ -16,7 +16,7 @@ const tableNumberStyle = css`
     padding-right: 5em;
 `;
 
-export function QueryPageNrOfIsolatesComponent(): JSX.Element {
+export function QueryPageNrOfIsolatesComponent(props: {nrOfSelectedIsol: number}): JSX.Element {
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Remove Context from view "

src/app/Main/Body/QueryPage/generateOneTableRowObj.service.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (
+            _.includes(uniqIsolateColValues, colElement) &&
+            _.includes(uniqIsolateRowValues, rowElement)
+        ) {
+            isolateCountGroups.forEach((isolateGroup) => {
+                if (
+                    (isolateGroup[rowAttribute] === rowElement ||
+                        isolateGroup[rowAttribute] === undefined) &&
+                    (isolateGroup[colAttribute] === colElement ||
+                        isolateGroup[colAttribute] === undefined)
```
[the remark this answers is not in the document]
MicheleK17: The final else-condition is, that the forEach-Loop of the isolateGroups should continue. If the Row/Col do not match, the count should not be extracted. 

I could write it that way if it helps. 

`if (
                    (isolateGroup[rowAttribute] === rowElement ||
                        isolateGroup[rowAttribute] === undefined) &&
                    isolateGroup[colAttribute] === colElement
                ) {
                    const statTableKey = isolateGroup[colAttribute];
                    tempStatTable[statTableKey] = String(isolateGroup.count);
                }
                if (
                    isolateGroup[rowAttribute] === rowElement &&
                    isolateGroup[colAttribute] === undefined
                ) {
                    const statTableKey = allValuesText;
                    tempStatTable[statTableKey] = String(isolateGroup.count);
                }`

src/app/Shared/LoadingOrError.component.tsx
MicheleK17: Done. Commit "Change if-conditions in LoadingOrError"

src/app/Shared/LoadingOrError.component.tsx
MicheleK17: Done. Commit "Change if-conditions in LoadingOrError"

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -114,95 +143,153 @@ export function QueryPageContainerComponent(): JSX.Element {
...
+    const fetchIsolateCounted = async (): Promise<void> => {
+        const isolateCountResponse: Response = await fetch(
+            isolateCountUrl
         );
-    }, [filter, table, isolateCountUrl]);
+        setIsolateCountStatus(isolateCountResponse.status);
+        if (isolateCountResponse.status === 200) {
+            const isolateCountProp: IsolateCountedDTO = await isolateCountResponse.json();
+            if (isolateCountProp.groups !== undefined) {
+                setTableContext(isolateCountProp);
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Remove useless if-undefined-check"

src/app/Main/Body/QueryPage/generateOneTableRowObj.service.ts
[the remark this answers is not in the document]
GerberI: yeah, that's better, i would put the col checks first in the if statements because that is what you set here.

GerberI (approved):
Commit: Rename handleChange-Functions, update TypeDocs
thx,
Now you have a lot of boiler plate code. But that is typical by using frameworks, you have a lot of boilerplate but the actual code is always the same. So actually you do not need to use your brain for writing that code. And that's where the editor comes in, if the boilerplate is always the same you can use code snippets or auto completion functionalities to write that boilerplate faster or run code checkers to validate the code.

Two reasons why your code always has the same handlers again and again:
- Your components are in general to small and fine grained. You often use components for e.g. one specific button. But to find the right balance is difficult and requires a lot of experience.
- As stated in comment below you often define your component by looking at the context there are used in. So you often have props down the tree that look how they look because you want them to use that way up in the tree. Typically if you use a component in another component you would convert the props of the used component so that they match the context of the using component. Then your handler functions would actually do some stuff, the mapping of the data to the relevant context. E.g. looking at a simplified radio button example:
```
function RadioButtonComponent(onClick: () => void){...}
function GenderDialogComponent(...) {
  return
    <RadioButtonComponent label="female" onClick="handleFemaleClick()"/>
    <RadioButtonComponent label="male" onClick="handleMaleClick()"/>
}
```
So the wrong way would be ignoring the dialog context and moving the responsibility handling the clicks up the tree:
```
const handleFemaleClick = () => props.onFemaleClick();
const handleMaleClick = () => props.onMaleClick();
```
The right way would be adapting the radio button context to the dialog context:
```
const handleFemaleClick = () => props.onGenderChange(Gender.Female)
const handleMaleClick = () => props.onGenderChange(Gender.Male)
```

Hope this was not to therotical :)

src/app/Core/ExportServices/generateExportLabels.service.ts
```diff
@@ -2,7 +2,7 @@ import { useTranslation } from "react-i18next";
 import { getCurrentDate } from "../getCurrentDate.service";
 import { MainFilterLabelInterface } from "../../Shared/Model/Export.model";
 
-export interface ExportLabels {
+export interface ExportLabelsInterface {
```
GerberI, line 6: better was ExportLabels, we had that a long time ago, adding Interface does not add any meaning to the name. You name interfaces like you would name the actual object it represents, so ExportLabels

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/Drawer-DisplFeatures.component.tsx
```diff
@@ -46,7 +46,7 @@ const iconStyle = css`
     height: 36px;
 `;
 
-interface DrawerDisplayedFeaturesInterface {
+interface DrawerDisplayedFeaturesProps {
```
GerberI, line 45: export

GerberI (approved):
Commit: Rename Interface => Prop

src/app/Main/Body/QueryPage/ChangeDisplFeatures.service.tsx
```diff
@@ -1,11 +1,10 @@
-import { ValueType } from "react-select";
 import { FilterType } from "../../../Shared/Model/Filter.model";
 
 export function handleChangeDisplayedFeatures(
-    selectedOption: ValueType<{value: string, label: string}>
+    selectedOption: {value: string, label: string}
 ): FilterType {
-    if (selectedOption) {
-        const selectedFeature: string = (Object.values(selectedOption)[0]);
+    if (selectedOption !== undefined && selectedOption.value !== undefined) {
```
GerberI: reject: selectedOption is typed and it says itself and the value cannot be undefined, so you just changed the type but not the values that can stand behind the type: see comments below for solution

src/app/Main/Body/QueryPage/ChangeFilter.service.ts
```diff
@@ -1,11 +1,9 @@
-import { ValueType } from "react-select";
-
 export function changeFilter(
-    selectedOption: ValueType<{value: string, label: string}>
+    selectedOption: { value: string; label: string }[]
 ): string[] {
-    if (Array.isArray(selectedOption)) {
+    if (selectedOption !== undefined) {
```
GerberI: same here

src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
```diff
@@ -46,7 +46,7 @@ export function SelectorListSelectorComponent(
     const handleChange = (
         selectedOption: ValueType<{ value: string; label: string }>,
         keyName: FilterType | TableType
-    ): void => props.onChange(selectedOption, keyName);
+    ): void => props.onChange(selectedOption as { value: string; label: string }[], keyName);
```
GerberI: same here: 
```
if(selectedOption !== undefined && Array.isArray(selectedOption) {
  props.onChange(...)
} else {
  throw error or just do nothing or
  props.onChange([])
}
```

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
```diff
@@ -74,7 +74,7 @@ export function DisplayedFeatureSelectorComponent(
     const handleChange = (
         selectedOption: ValueType<{ value: string; label: string }>,
         keyName: FilterType | TableType
-    ): void => props.onChange(selectedOption, keyName);
+    ): void => props.onChange(selectedOption as { value: string; label: string }, keyName);
```
GerberI: reject: see comment above: you should do type checking here, something like
```
if(selectedOption !== undefined && !Array.isArray(selectedOption) {
  props.onChange(...)
} else {
  //  depends on what selection is allowed and is a reasonable output for this component
  throw error or just do nothing or
  props.onChange({value: "", label: ""}) or
  props.onChange(null)
}
```
So you check if the ValueType looks like you expect it to look like.

GerberI (changes requested):
Commit: Adapt types of selector onChange
Only half of the work is done...

GerberI (approved):
Commit: Rename services choose selected values
Good,
So this was the same issue as stated in a previous commit: Before the services were named after the context they were used in, now they are named after the context they provide themselves.

src/app/Main/Body/QueryPage/chooseSelectedDisplFeatures.service.ts
```diff
@@ -1,9 +1,9 @@
 import { FilterType } from "../../../Shared/Model/Filter.model";
 
 export function chooseSelectedDisplayedFeaturesService(
-    selectedOption: {value: string, label: string}
+    selectedOption: { value: string; label: string } | null
 ): FilterType {
-    if (selectedOption !== undefined && selectedOption.value !== undefined) {
+    if (selectedOption && selectedOption.value) {
```
GerberI: General consideration: prefer always using the exact value you're checking against. E.g. `if(selectedOption.value)` is different from `if(selectedOption.value !== undefined)`. The first check will be false for `value = undefined | null | ""`

GerberI (approved):
Commit: Fix bug - handle remove last selected value
I wonder if you actually need the label when the selection changes? Would it not be sufficient to use something like
- `onSingleSelectionChange(value: string)` => value = "" means nothing selected
- `onMultiSelectionChange(value: string[])` => value = [] means nothing selected

src/app/Main/Body/QueryPage/dataApi.service.ts
```diff
@@ -0,0 +1,41 @@
+import { adaptIsolatesFromAPI } from "../../../Core/adaptIsolatesFromAPI.service";
+import { FilterConfigDTO } from "../../../Shared/Model/Api_Filter.model";
+import { IsolateDTO } from "../../../Shared/Model/Api_Isolate.model";
+import { DbCollection } from "../../../Shared/Model/Client_Isolate.model";
+import { FilterInterface } from "../../../Shared/Model/Filter.model";
+import { generateUniqueValuesService } from "./generateUniqueValues.service";
+
+export async function dataApiService(
```
GerberI: Recommendations
- don't merge the different api endpoints
- i would use the urls right here, no need to use them as a parameter
- use a generic type for the response:
```
interface ApiResponse<T extends {}>{
  status: number,
  data?: T
}
```
- as for now the api services do not really do something useful, you've could have placed the stuff to your mapping services and just renamed them: adaptIsolatesFromApi => fetchIsolatesFromApi. If you extract the fetching and jsoning to a general api service this would become more obvious.

GerberI (approved):
Commit: Extract api-code to service

GerberI (approved):
Commit: Reorganize and Rename
Good, now your services are located with the components using them.
I wonder why you name the service folders like QueryPageServices, they are already in a folder QueryPage so its a duplicate naming, just Services would be fine.

GerberI (approved):
Commit: Remove Context from view
thx,
The "clutter" in the props will become less if you make the props more sensitive to the context of the components they are used for (as stated in a previous review). This was not possible until now because you used the same react context data for each component. I recommend to check and change the props of a component when it is affected by a future ticket, so that your code slowly progresses to a more cleaner form.
And if your layout components become real layout components a lot of the props will become obsolete.

GerberI (approved):
Commit: Change if-conditions in LoadingOrError
thx

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -230,10 +230,7 @@ export function QueryPageContainerComponent(): JSX.Element {
 
         setIsolateCountStatus(tableResponse.isolateCountStatus);
 
-        if (
-            tableResponse.isolateCountGroups !== undefined &&
-            tableResponse.nrOfSelectedIsolates !== undefined
-        ) {
+        if (tableResponse.nrOfSelectedIsolates !== undefined) {
```
GerberI, line 233: it is now the same check as below, you can merge them

GerberI (approved):
Commit: Remove useless if-undefined-check

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -191,7 +191,7 @@ export function QueryPageContainerComponent(): JSX.Element {
 
             const allValuesText = t("Results.TableHead");
 
-            const colValues = generateTableHeaderValuesService(
+            const columnNames = generateTableHeaderValuesService(
```
GerberI, line 194: General consideration: Prefer your state variables over your temporary variables: So columnNames for state and e.g. colNames for the temp variable.
This is one more reason why i am no fan of the functional react components, in a class there would be a clear difference between state and local variable: state => this.columnNames, local => columnNames

src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
```diff
@@ -102,16 +98,12 @@ export function QueryPageLayoutComponent(props: {
...
-                colAttributes={props.colAttributes}
-                tableColumn={props.tableColumn}
-                tableRow={props.tableRow}
-                tableOption={props.tableOption}
-                tables={{
-                    statisticDataAbsolute: props.tables.statisticDataAbsolute,
-                    statisticDataPercent: props.tables.statisticDataPercent
+                columnNameValues={props.columnNameValues}
+                tableContext={props.tableContext}
+                numberOfIsolates={{
```
GerberI, line 103: ? numberOfIsolates = {props.numberOfIsolates}

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/calculateRelativeTableData.service.ts
```diff
@@ -1,10 +1,10 @@
...
-    const percentRounded = percent.toFixed(2);
-    return percentRounded;
+function calculateRelative(part: number, total: number): string {
+    const relative = total === 0 ? 0 : (part / total) * 100;
+    const relativeRounded = relative.toFixed(2);
+    return relativeRounded;
 }
 /**
- * @desc Calculate the percentage of an objectList with absolute numbers
+ * @desc Calculate the relativeage of an objectList with absolute numbers
```
GerberI, line 7: "Calculate the relative numbers ..." perhaps?

src/app/Main/Body/QueryPage/Results/TableResults/QueryPageContent-TableResultsLayout.component.tsx
```diff
@@ -30,14 +30,8 @@ export interface TableResultsProps {
...
-    columnAttributes: string[];
-    tableColumn: string;
-    tableRow: string;
-    tableOption: DisplayOptionType;
-    tables: {
-        statisticDataAbsolute: Record<string, string>[];
-        statisticDataPercent: Record<string, string>[];
-    };
+    columnNameValues: string[];
+    tableContext: TableInterface
```
GerberI, line 34: Consider not using context as a name for a prop (because it is a reserved react keyword), just table or tableData would be fine.

src/app/Main/Body/QueryPage/Results/TableResults/TableResults-Options.component.tsx
```diff
@@ -52,7 +52,7 @@ const BlueRadio = withStyles(() =>
 )(Radio);
 
 /**
- * @desc Returns the option bar to display the table numbers as absolute numbers or in percent
+ * @desc Returns the option bar to display the table numbers as absolute numbers or in relative
```
GerberI, line 55: as absolute or relative numbers?

src/app/Main/Body/QueryPage/Results/TableResults/TableResults-Options.component.tsx
```diff
@@ -82,9 +82,9 @@ export function ResultsTableOptionsComponent(props: {
                     />
                     <FormControlLabel
                         css={radioButtonSizeStyle}
-                        value="percent"
+                        value="relative"
                         control={<BlueRadio color="default" size="small" />}
-                        label="Percent"
+                        label="Percentage"
```
GerberI, line 87: translation missing (i think this is already a ticket)

GerberI (approved):
Commit: Renaming & group props of QueryPage-Components
General consideration: There is a good convention for naming multiples of things: adding an "s". E.g. columnNameValues would be columnNames, same for e.g. columnNameArray => columnNames or columnNameList => columnNames. If you stick to that your loops would become very self explaining: columnNames.foreach(columnName => ...)

GerberI:
So if that's it, and this was a lot, i would change the ticket name to Prepare Changing Container-View-Structure :) Because the structure has not really changed. So this refactoring should have been splitted into multiple tickets:
- Remove context from view components
- Combine query page containers
- Move feature dependent services from core
- and so on ...
So as a next step in this epic refactoring i would propose to convert the pseudo layouts to real layout components. This would remove a lot of duplicated code and make the overall component structure more clear.
Later steps could be:
- Make the view components more generic and clean, merge some of them to a more compact form
- Extract the then clean view components that are generic and reusable to the shared module
- Refactor the errorPage to be more generic and data independent and to act as a view not a layout
- Extract the whole QueryPage to a new main feature to separate it from the common page stuff
- ...
Some of that refactoring i would do during implementing tickets touching the components. I would write that in detail in the ticket descriptions so that you have an idea how the component should look like after the refactoring. And before you do further refactoring regarding the component tree i would create a component tree diagram, so that you know beforehand what you have to do.

src/app/Main/Body/QueryPage/ChangeDisplFeatures.service.tsx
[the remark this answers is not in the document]
GerberI: partly done: just converting the type does not do the actual check if the type is the one you expect, see later review

src/app/Main/Body/QueryPage/ChangeFilter.service.ts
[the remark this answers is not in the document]
GerberI: obsolete

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
```diff
@@ -36,6 +39,17 @@ const useStyles = makeStyles(() =>
...
+        selectedOption: ValueType<{value: string, label: string}>,
+        keyName: FilterType | TableType
+    ) => void;
+    handleSwapDisplFeatures: () => void;
+    handleChangeFilter: (
+        selectedOption: ValueType<{value: string, label: string}>,
+        keyName: FilterType | TableType
+    ) => void;
+    handleRemoveAllFilter: () => void;
+    handleRemoveAllDisplFeatures: () => void;
```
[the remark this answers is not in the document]
GerberI: move this to a new layout refactoring ticket

src/app/Main/Body/QueryPage/generateOneTableRowObj.service.ts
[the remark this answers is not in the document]
MicheleK17: Done. Commit "Change generateOneTableRowObj.service"

src/app/Main/Body/QueryPage/Drawer/ControlBar/DrawerControl-ResizeBar.component.tsx
```diff
@@ -30,18 +30,20 @@ export interface ResizeBarProps {
...
- * @param {(newWidth: number) => void} onChange - onChange-function for changing the size of the Drawer
+ * @param {ResizeBarProps} props - onChange-function for changing the size of the Drawer
  * @returns {JSX.Element} - resize bar component
  */
 export function DrawerControlResizeBarComponent(
     props: ResizeBarProps
 ): JSX.Element {
     const minDrawerWidth = 325;
 
+    const handleChange = (newWith: number): void => props.onChange(newWith);
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit: "Fix typo & Remove types"

src/app/Main/Body/LinkPage/ListContent-ListItem.component.tsx
```diff
@@ -9,8 +9,7 @@ export interface ListItemProps {
 
 /**
  * @desc Returns one ListItem for a link
- * @param {string} link - link URL
- * @param {string} text - text to describe the link
+ * @param {ListItemProps} props
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit: "Fix typo & Remove types"

src/app/Core/ExportServices/generateExportLabels.service.ts
MicheleK17: Done. Commit: "Remove "Interface"-Label"

src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
MicheleK17: Done. Commit "Little Changes"

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: Done. Commit "Little Changes"

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/Drawer-DisplFeatures.component.tsx
MicheleK17: Done. Commit "Little Changes"

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/calculateRelativeTableData.service.ts
MicheleK17: Done. Commit "Little Changes"

src/app/Main/Body/QueryPage/Results/TableResults/QueryPageContent-TableResultsLayout.component.tsx
MicheleK17: Done. Commit "Little Changes"

src/app/Main/Body/QueryPage/Results/TableResults/TableResults-Options.component.tsx
MicheleK17: Done. Commit "Little Changes"

src/app/Main/Body/QueryPage/ChangeDisplFeatures.service.tsx
MicheleK17: This function has been changed to export function chooseSelectedDisplayedFeaturesService.  Can you please look into these functions again and say if changes are still needed?

src/app/Main/Body/QueryPage/ChangeFilter.service.ts
MicheleK17: This function has been changed to export function export function chooseSelectedFiltersService. Can you please look into these functions again and say if changes are still needed?

src/app/Main/Body/QueryPage/Drawer/Displayed_Features/DisplFeatures-Selector.component.tsx
MicheleK17: Done. Commit "Type validation in selector functions"

src/app/Main/Body/QueryPage/Drawer/Filter/SelectorList-Selector.component.tsx
MicheleK17: Done. Commit "Type validation in selector functions"

src/app/Main/Body/QueryPage/Drawer/ControlBar/QueryPage-DrawerControl.component.tsx
```diff
@@ -59,18 +59,21 @@ const drawerIconStyle = css`
...
+     */
     isOpen: boolean;
+     /**
+     * width of the Drawer after resizing
+     */
     newWidth: number;
-    handleDrawer: () => void;
-    handleResize: (newWidth: number) => void;
+    onDrawerOpenCloseClick: () => void;
+    onDrawerSizeChange: (newWidth: number) => void;
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit: "Change drawer resize naming"

src/app/Main/Body/QueryPage/Drawer/ControlBar/QueryPage-DrawerControl.component.tsx
```diff
@@ -59,18 +59,21 @@ const drawerIconStyle = css`
 `;
 
 export interface DrawerControlProps {
+    /**
+     * true if Drawer is open
+     */
     isOpen: boolean;
+     /**
+     * width of the Drawer after resizing
```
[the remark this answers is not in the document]
MicheleK17: Done. Commit: "Change drawer resize naming"

src/app/Main/Body/QueryPage/dataApi.service.ts
MicheleK17: Improved in commit "Change ApiServices"

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
```diff
@@ -13,29 +13,45 @@ export function generateOneTableRowObjService(
...
-                    tempStatTable[statTableKey] = String(isolateGroup.count);
+                if (breakIsolateCountLoop) {
+                    return;
+                }
+
+                const isolateGroupRowValue = isolateGroup[rowAttribute];
+                const isolateGroupColValue = isolateGroup[colAttribute];
+                const isolateCount = isolateGroup.count;
+
+                const onlyRowNameIsInIsolateGroup: boolean =
```
GerberI, line 32: Reject: The algorithm is still problematic: E.g. if `ig[row] === rowName && ig[col] === undefined` comes first in the isolateGroups the loop ends always in setting the allValues entry and the other cols will be ignored. So a solution would be to first scan through the isolateGroups in the loop and set booleans for things that where found during the loop. After the loop you can then check these booleans and decide what to do.
As stated it would be good to convert the isolateCount data from the API to a more convenient format  before doing anything with it. I would create a new ticket for that and if you like we can develop a conversion  algorithm together.

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
```diff
@@ -13,29 +13,45 @@ export function generateOneTableRowObjService(
...
-        if (
+    columnNameValues.forEach((colName: string) => {
+        const isolateGroupIncudesRowAndCol: boolean =
             _.includes(uniqIsolateColValues, colName) &&
-            _.includes(uniqIsolateRowValues, rowName)
-        ) {
+            _.includes(uniqIsolateRowValues, rowName);
+
+        if (isolateGroupIncudesRowAndCol) {
+            let breakIsolateCountLoop = false;
```
GerberI, line 22: (obsolete due to reject below)
unnecessary, you could return directly, see below

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
```diff
@@ -13,29 +13,45 @@ export function generateOneTableRowObjService(
...
+                    isolateGroupColValue === undefined;
+                const colNameIsInIsolateGroup: boolean =
+                    isolateGroupColValue === colName &&
+                    (isolateGroupRowValue === rowName ||
+                        isolateGroupRowValue === undefined);
+
+                if (colNameIsInIsolateGroup) {
+                    const statTableKey = isolateGroupColValue;
+                    tempStatTable[statTableKey] = String(isolateCount);
+                    breakIsolateCountLoop = true;
```
GerberI, line 43: return;

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
```diff
@@ -13,29 +13,45 @@ export function generateOneTableRowObjService(
...
+                } else if (onlyRowNameIsInIsolateGroup) {
+                    const statTableKey = allValuesText;
+                    tempStatTable[statTableKey] = String(isolateCount);
+                    breakIsolateCountLoop = true;
```
GerberI, line 47: return;

GerberI (changes requested):
Commit: Change generateOneTableRowObj.service
Good, you have introduced new variables to make the code more readable. But because it becomes more readable it is easier to find bugs ...

GerberI (approved):
Commit: Fix typo & Remove types
Thx, now the comments are also easier to read. I hope it was not to frustrating to undo all the work you've done.
Hint: You can also remove the typings from the return comments:
`@returns {string} - modified string` => `@returns modified string`
`@returns {void}` => remove completely, it returns nothing

GerberI (approved):
Commit: Remove "Interface"-Label
thx

GerberI (approved):
Commit: Little Changes
thx

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -114,95 +143,153 @@ export function QueryPageContainerComponent(): JSX.Element {
...
+                statisticDataAbsolute: [],
+            });
+        } else {
+            const adaptedIsolateCounts: ClientIsolateCounted = {
+                totalNumberOfIsolates: isolateCountProp.totalNumberOfIsolates,
+                groups: adaptCountedIsolatesGroupsService(
+                    isolateCountProp.groups
+                ),
+            };
+
```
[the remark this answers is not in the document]
GerberI: => new ticket

.eslintrc.js
```diff
@@ -66,5 +66,8 @@ module.exports = {
         "@typescript-eslint/no-explicit-any": 2,
         // Not too fond of this one
         "unicorn/filename-case": "off",
+
+        "unicorn/no-null": "off",
```
GerberI, line 70: 👍

GerberI (approved):
Commit: Type validation in selector functions
Good, now the selector components produce a well defined output and components using them can rely on the output as defined in the props interface.

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
```diff
@@ -68,7 +68,7 @@ export interface DrawerLayoutProps {
  * @returns {JSX.Element} - Drawer component
  */
 export function DrawerLayoutComponent(props: DrawerLayoutProps): JSX.Element {
-    const classes = useStyles((props.newWidth as unknown) as string);
+    const classes = useStyles((props.drawerWidth as unknown) as string);
```
GerberI, line 71: old story: useStyles(String(drawerWidth))

GerberI (approved):
Commit: Change drawer resize naming
thx

src/app/Core/api.service.ts
```diff
@@ -0,0 +1,20 @@
+export interface ApiResponse<T>{
+    apiStatus: number;
+    apiProp?: T;
+}
+
+export async function ApiService(apiUrl: string): Promise<ApiResponse<Record<string, unknown>[]>> {
```
GerberI, line 6: Sorry, i should have explained it in more detail. You can use the generic type T for functions too:
`function ApiService<T extends Object>(url: string): Promise<ApiResponse<T>>`
So no conversions necessary on calling the service: `const foo: FooDTO = ApiService('fooUrl)`
I think `Object` is the more appropriate type for a generic json response.

src/app/Core/api.service.ts
```diff
@@ -0,0 +1,20 @@
...
+    apiProp?: T;
+}
+
+export async function ApiService(apiUrl: string): Promise<ApiResponse<Record<string, unknown>[]>> {
+    const apiResponse: Response = await fetch(apiUrl);
+
+    const apiStatus = apiResponse.status;
+
+    if (apiStatus === 200) {
+        const apiProp: Record<string, unknown>[] = await apiResponse.json();
```
GerberI, line 12: So here it would be: `const apiProp: T = await apiResponse.json()`

src/app/Core/api.service.ts
```diff
@@ -0,0 +1,20 @@
+export interface ApiResponse<T>{
```
GerberI, line 1: this should also be typed: `interface ApiResponse<T extends Object>`

src/app/Core/api.service.ts
```diff
@@ -0,0 +1,20 @@
+export interface ApiResponse<T>{
+    apiStatus: number;
+    apiProp?: T;
```
GerberI, line 3: so you are in the api interface i would use just status and data as names.

src/app/Core/api.service.ts
GerberI, line 6: Consider naming your service as functions because your services are functions, e.g.:
`callApiService()` or `fetchJsonFromApi()`
Nouns like `ApiService` indicate a class like usage: `ApiService.fetch()`
So i would suggest for later tickets to use classes for Services, then you can also encapsulate multiple functions in one service class:
```
class ApiService {
  public static fetchFromUrl(url: string) {...}
  public static fetchFromFile(fileName: string) {...}
}
const foo = ApiService.fetchFromUrl('fooUrl');
```

GerberI (changes requested):
Commit: Change ApiServices
Thx,
sorry for the confusion, so there is for now no additional value of having separate api services you are right to undo the changes and remove them.
I would suggest to add the minor changes regarding the generic types as commented to the ApiService.

src/app/Main/Body/QueryPage/ChangeDisplFeatures.service.tsx
GerberI: looks good, but service is kind of obsolete now and can be replaced by:
`selectedFeature = selectedOption  === null ? "" : selectedOption.value`
or completely removed if your selector component returns: `onSelectionChange(value: string)` with value = "" for no selection.

src/app/Main/Body/QueryPage/ChangeFilter.service.ts
GerberI: looks good, but service is kind of obsolete now and can be replaced by:
`selectedFilters = selectedOption  === null ? [] : selectedOption.map(option => option.value)`
or completely removed if your selector component returns: `onSelectionChange(options: string[])` with options = [] for no selection.

src/app/Main/Body/QueryPage/ChangeDisplFeatures.service.tsx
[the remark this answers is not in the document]
GerberI: Done

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
MicheleK17: Of cause, I tried this solution first. But the linter says "Unnecessary return statement." as an error.

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
MicheleK17: Of cause, I tried this solution first. But in this case I get the error "Uncaught TypeError: rule is undefined" of the browser.

src/app/Core/api.service.ts
MicheleK17: Yes, I understood that. But when I do it like this I get the error "can not find name T" in the return type Promise<ApiResponse<T>>.

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
GerberI: Sorry, my fault. The linter is right. Ignore my "return" comments. I've ignored that we are in a for each loop and a return does not cancel the loop but only one iteration of the loop, so you have to use the boolean to cancel each following loops.

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
GerberI: wrong, you're right

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
GerberI: wrong, you're right

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
GerberI: Seems to be a problem with your input "drawerWidth". Please log the value to the console and check if it is always a number. I suspect that it is at some initial point undefined or null, then you should set it to a default for the useStyles.
Or you have to add a parameter to the useStyles function? => `useStyles = makeStyles(drawerWidth: string)`

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
MicheleK17: No, the other columns are not ignored. The col-for-each loop is independent of the boolean break. The break is only for the ig iteration that starts for each column. 
But yes, of course we can rethink this alogrithm in a later ticket.

src/app/Core/api.service.ts
GerberI: ? This works:
```
interface ApiResponse<T extends Object> {
    status: number;
    prop?: T
}
async function getFromApi<T extends Object>():Promise<ApiResponse<T>> {
    const response = await fetch('someUrl');
    const x: ApiResponse<T> = {
        status: response.status,
        prop: await response.json()
    }
    return x;
}

interface MyData {
    foo: number;
}

async function test() {
    const response: ApiResponse<MyData> = await getFromApi();
    if(response.prop !== undefined) {
        const foo = response.prop.foo; // typed as number
    }
}
```

GerberI: Hint: you can also use a shorter form: `const response = await getFromApi<MyData>();` in the test function.

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
GerberI: Re-Comment (this comes first) (messed up with github review, sorry): Yes, but for each col-for-each the ig loop will success on the row = rowName and col = undefined check and set the allValuesText in my example situtation. Then the ig loop cancelling boolean is set and all other igs are ignored.

src/app/Core/api.service.ts
MicheleK17: Ah I forgot the Type behind the function. But I have to use the type Record<string, unknown>[]  instead of Object, because its a more precise description. (linter)

MicheleK17: Okay. I can not use Record<string, unknown>, because its not matching the DTO. But when I use Object, the linter is complaining. Should I deactivate this rule?

GerberI: Perhaps Record<string, any>? or T extends any would also be possible (that's what comes from the json() parser.

GerberI: Hint: You can disable a linter rule for a specific line

GerberI: What does the linter say to Object? You can also use T extends {} | [{}]

MicheleK17: Can I remove the "extends X" completely? than it would work. 
Using any is also forbitten by the linter. 
Disable a linter rule for one line seems so hacky ?! 😄

MicheleK17: linter to Object: Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.

GerberI: Disabling a linter rule for a line is not hacky if you know what you do. Your json() gives you any as a result so you know that and you can disable the linter rule to adapt to it.

GerberI: So the linter is only for hinting you that something in your code might be wrong, which is in most cases true.

GerberI: I would not remove the extend, only T means unknown, but you know that your api has a specific format: Object, so you should type it accordingly.

MicheleK17: And the linter complains about {} in the same way.

GerberI: So, i would prefer {} and disable the rule for the relevant lines. If your api returns an array not encapsulated in an object (i don't know) it would be {} | [{}]

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
MicheleK17: Done.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -136,15 +146,25 @@ export function QueryPageContainerComponent(): JSX.Element {
...
-        const dataResponse = await dataApiService(ISOLATE_URL, FILTER_URL);
+        const isolateResponse = await ApiService(ISOLATE_URL);
+        const filterResponse = await ApiService(FILTER_URL);
+
+        setIsolateStatus(isolateResponse.apiStatus);
+        setFilterStatus(filterResponse.apiStatus);
 
-        setIsolateStatus(dataResponse.isolateStatus);
-        setFilterStatus(dataResponse.filterStatus);
+        if (isolateResponse.apiProp && isolateResponse.apiProp) {
```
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Core/api.service.ts
MicheleK17: Done.

src/app/Core/api.service.ts
MicheleK17: Done.

src/app/Core/api.service.ts
MicheleK17: Done.

src/app/Core/api.service.ts
MicheleK17: Done.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -136,15 +146,25 @@ export function QueryPageContainerComponent(): JSX.Element {
...
+            const isolateProp = (isolateResponse.apiProp as unknown) as IsolateDTO;
+            const filterProp = (filterResponse.apiProp as unknown) as FilterConfigDTO;
```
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -226,16 +246,16 @@ export function QueryPageContainerComponent(): JSX.Element {
...
+        setIsolateCountStatus(tableResponse.apiStatus);
 
-        if (tableResponse.nrOfSelectedIsolates !== undefined) {
-            setTableContext(
-                tableResponse.isolateCountGroups,
-                tableResponse.nrOfSelectedIsolates
-            );
-            setNrOfSelectedIsol(tableResponse.nrOfSelectedIsolates)
+        if (tableResponse.apiProp !== undefined) {
+            const isolateCountProp = (tableResponse.apiProp as unknown) as IsolateCountedDTO;
```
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Core/api.service.ts
MicheleK17: Renameing done.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -31,19 +32,20 @@ export function HeaderExportContainerComponent(): JSX.Element {
...
 
         let rawData: DbCollection = [];
         let rawKeys: DbKey[] = [];
         let statData: Record<string, string>[] = [];
         let statKeys: string[] = [];
 
-        if (isolateFilteredStatus === 200) {
-            const isolateFilteredProp: IsolateDTO = await isolateFilteredResponse.json();
+        if (isolateFilteredStatus === 200 &&isolateFilteredResponse.apiProp !== undefined) {
+            const isolateFilteredProp = isolateFilteredResponse.apiProp as unknown as IsolateDTO;
```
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
MicheleK17: I don't get it. I already have a default for the width. The width is always of type number. Even if I check the type before converting it doesn't work. I also tried to convert the type directly in the handler and not in the Layout. But that does not work either.

GerberI: Strange, you can leave it as it is because i have to test it myself to help further. Perhaps i will check it when you merged to develop.

GerberI (approved):
Commit: Some more changes
You fixed the width to string problem!!!
Some general consideration as always (refresh):
Because  programmers are lazy as default here my convention to when to type a variable:
`const x = 5` => not necessary, type inferred from `5` statement
`const x = foo()` => not necessary, type inferred from return type of function
`const x = y.foreach(...)` => not necessary, type inferred from foreach function
`const x: { foo: number } = { foo: 5 }` => useful because you can define something wrong
`const x: number = fooReturnsAny()` => same as `const x = fooReturnsAny() as number`
So why you can leave the type in some situations is that the compiler will complain when you use the variable: `bar(x)`. In contrast to the compiler complaining when the variable is defined and typed: `const x: Type = foo()`
So what you really need to type are function parameters and return types when you define a new function, interfaces per definition and when you define new objects. Also class members if you use classes.
E.g. even C# as a very strongly typed language has introduced the `var` keyword for local variables to prevent bugs and make code more readable: Instead of using `Type x = foo()` they recommend to use `var x = foo()` to let the compiler infer the correct type.

src/app/Main/Body/QueryPage/QueryPageServices/TableServices/generateOneTableRowObj.service.ts
MicheleK17: According to api, there is only one associated isolate group per field. That means, when I have found it, I can cancel the loop over the ig, because I won't find any in the following ones anyway. Sure it depends on the API, but I don't think it's that buggy.

GerberI: Obsolete, due to misunderstanding, the api response has a different semantic for the discussed problematic cases.

PR #246, opened 2021-04-16
Limit height of selector dropdown

MicheleK17:
Ticket: #218
Change-Id: I9c21e027baf9fa4afe58566cbf3a1d004bff52c2


GerberI (approved):
Would be better if there would be a prop visibleItemCount or so in the Selector lib, but i think there is none.
You could add one to your component and then calculate the height with the margins and font size. You can consider this if you happen to need different visible items for different lists in your app.

Issue #245, opened 2021-04-16
As a User I would like to be able to contact zoonotify-support via the 'Problem Melden' button


dptolle:
@MicheleK17 commented on [Fri Jul 24 2020](https://github.com/SiLeBAT/zoonotify-server/issues/15)

Retrieve the support-email address from the /info endpoint and add it into the relevant HTML Element

PR #249, opened 2021-04-19
Use support mail from api

MicheleK17:
Ticket: #245
Change-Id: I419138be7e6112f1ce7c7053a7ed44a4fa9f834a

### DIFF
--- a/src/app/Main/Footer/Footer-Layout.component.tsx
+++ b/src/app/Main/Footer/Footer-Layout.component.tsx
@@ -1,7 +1,5 @@
 /** @jsx jsx */
 import { css, jsx } from "@emotion/core";
-import { FooterLinkListComponent } from "./Footer-LinkList.component";
-import { LastUpdateComponent } from "./LastUpdate.component";
 import {
     surfaceColor,
     primaryColor,
@@ -18,12 +16,13 @@ const footerStyle = css`
 `;
 
 export function FooterLayoutComponent(props: {
-    supportMail: string | undefined;
+    lastUpdateComponent: JSX.Element;
+    linkListComponent: JSX.Element;
 }): JSX.Element {
     return (
         <footer css={footerStyle}>
-            <LastUpdateComponent />
-            <FooterLinkListComponent supportMail={props.supportMail} />
+            {props.lastUpdateComponent}
+            {props.linkListComponent}
         </footer>
     );
 }
--- a/src/app/Main/Footer/Footer-Layout.component.tsx
+++ b/src/app/Main/Footer/Footer-Layout.component.tsx
@@ -17,11 +17,13 @@ const footerStyle = css`
     border-top: 2px solid ${primaryColor};
 `;
 
-export function FooterLayoutComponent(): JSX.Element {
+export function FooterLayoutComponent(props: {
+    supportMail: string | undefined;
+}): JSX.Element {
     return (
         <footer css={footerStyle}>
             <LastUpdateComponent />
-            <FooterLinkListComponent />
+            <FooterLinkListComponent supportMail={props.supportMail} />
         </footer>
     );
 }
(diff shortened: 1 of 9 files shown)


src/app/Main/Footer/Footer-LinkList.component.tsx
```diff
@@ -41,7 +41,9 @@ const linkStyle = css`
     }
 `;
 
-export function FooterLinkListComponent(): JSX.Element {
+export function FooterLinkListComponent(props: {
+    supportMail: string | undefined;
```
GerberI, line 45: Reject: What happens if supportMail is undefined? Perhaps you could check it in the container and display an error message or disable the link here and show a tooltip with an error message. Or fetch the mail in the main container with the other stuff, use the error mechanism there and put the serverInfo into a context object used by the footer container.
You can also add a ticket if this would be to time consuming and it was not defined in the "support mail from api" ticket.

src/app/Main/Footer/Footer-Layout.component.tsx
```diff
@@ -17,11 +17,13 @@ const footerStyle = css`
     border-top: 2px solid ${primaryColor};
 `;
 
-export function FooterLayoutComponent(): JSX.Element {
+export function FooterLayoutComponent(props: {
```
GerberI, line 20: Reminder: This is no layout component for now, it would be if it would look like:
```
export function FooterLayoutComponent(serverInfoComponent: JSX.Element, linkListComponent: JSX.Element`) {
  return (
    <footer>
      <serverInfoComponent>
      <linkListComponent>
    </footer>
  );
}
```
And in the FooterContainer:
```
// sorry don't now how it would look in react, so pseudo code
return <FooterLayoutComponent 
  serverInfoComponent = <LastUpdateComponent>
  linkListComponent = <FooterLinkListComponent supportMail={props.supportMail}>
  />
```

src/app/Main/Footer/Footer-LinkList.component.tsx
MicheleK17: I think it is not the right way not to show the complete page just because the SupportMail is not delivered by the server (error mechanism of the QueryPage). This seems to be overkill. Blocking the link seems more appropriate to me.

MicheleK17: Done. Commit "Handle SupportMail is undefined"

src/app/Main/Footer/Footer-Layout.component.tsx
MicheleK17: Done. Commit "Refactor FooterLayout"

src/app/Shared/URLs.ts
```diff
@@ -6,6 +6,7 @@ export const ZNPaths = {
     dpdPagePath: "/dataProtectionDeclaration",
 };
 const VERSION_ROOT = "/v1";
+export const INFO_URL = `${VERSION_ROOT}/info`;
```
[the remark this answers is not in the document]
MicheleK17: I think I'll leave it as it is for now. There are a lot more than just components in "shared" at the moment.

GerberI (approved):
Commit: Handle SupportMail is undefined
Good, thx.

src/app/Shared/URLs.ts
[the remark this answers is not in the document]
GerberI: that's fine

src/app/Main/Footer/Footer-LinkList.component.tsx
GerberI: Yes, that's a difficult decision, how user friendly or error tolerant the app should be. For the mibi-portal i came to the solution to show a loading spinner blocking the whole page until all necessary data is fetched. But that's not implemented yet, the mibi-portal for now just looks normal, even if the api is not working. That was some kind of a problem during a release, we thought all was fine but then users complained because the api was not running...

GerberI (approved):
Commit: Refactor FooterLayout
cool, thx.
I thought this would be a good place to begin refactoring the layouts. I would do that whenever you happen to touch such a component in a ticket.

GerberI:
all done, ready to merge

Issue #259, opened 2021-05-14
There is a call (fetch) to the isolate endpoint when loading the Welcome Page and two when loading the QueryPage


dptolle:
This should not happen.  There's no need to load all the isolates when reaching the Welcome page.

There are two inappropriate API calls:
1) Header-ExportContainer.component.useEffect -> Calls the /isolate endpoint upon first render
2) QueryPage-Container.component.useEffect -> Calls the /isolate endpoint upon first render

This should not happen.  The /isolate endpoint downloads ALL isolate data (if no filter is set) and should only be called once the user decides to download the dataset.

PR #267, opened 2021-05-25
Improve performance

MicheleK17:
Ticket: #259
Change-Id: I59ef967becd2cc3739ae29ac5f9c0320ec06a183


src/app/App.tsx
```diff
@@ -1,20 +1,20 @@
 import React, { Suspense } from "react";
 import * as ReactDOM from "react-dom";
 import { StylesProvider } from "@material-ui/core/styles";
-import { CheckServerConnectionComponent } from "./Main/CheckServerConnection.component";
```
GerberI, line 4: I think not that checking the server connection in general is useless. As discussed there should be a mechanism connected to the loading of necessary data from the server on page load, that shows a spinner and/or displays a message when the loading failed. That is because of loading the web page and accessing the api are different mechanism or parts of the server and if the api part fails the user should be informed as soon as possible about that. So i think you already have that mechanism.

GerberI (approved):
Commit: Remove useless check of server connection

src/app/App.tsx
MicheleK17: But for this situation, I have the error page, right?

src/app/Main/Header/Export/Dialog/Header-ExportDialog.component.tsx
```diff
@@ -53,14 +54,14 @@ export function HeaderExportDialogComponent(props: {
     );
 
     return DialogComponent({
-        isOpen: props.isOpen,
+        isOpen: true,
```
GerberI, line 57: might be the time for removing the isOpen prop from the DialogComponent too

src/app/Main/Header/Header.component.tsx
```diff
@@ -63,15 +63,6 @@ const rightHeaderStyle = css`
     display: flex;
     align-items: flex-end;
 `;
-const subheaderStyle = (open: boolean): SerializedStyles => css`
```
GerberI, line 74: Reject: This style belongs to this component because it is for aligning the header elements (just one for now). So please leave it here. You could extract the export button specific part to the HeaderExportButtonComponent (not the HeaderExportContainer).

src/app/Main/Header/Header.component.tsx
```diff
@@ -134,9 +125,7 @@ export function HeaderComponent(): JSX.Element {
                 </div>
             </div>
 
-            <div css={subheaderStyle(queryOpen)}>
-                <HeaderExportContainerComponent />
-            </div>
+            {queryOpen && <HeaderExportContainerComponent />}
```
GerberI, line 128: Good, generally avoid display:none for putting together the UI. You use the react framework and it provides a flexible and fast way for rendering components depending on the apps state. So by using display:none you actually bypass the framework.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -20,10 +22,22 @@ import { generateExportLabels } from "./ExportServices/generateExportLabels.serv
...
+
+const subheaderStyle = css`
+    width: 100%;
+    display: flex;
+    justify-content: flex-end;
+    align-items: center;
+    background-color: ${bfrPrimaryPalette[300]};
+    box-sizing: border-box;
+    box-shadow: 0 8px 6px -6px grey;
+`;
```
GerberI, line 35: Reject: See other comment. In general a container should not contain any styles. Styles belong to the view and layout components, the container is for handling app logic and putting together views and layouts.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -88,13 +102,10 @@ export function HeaderExportContainerComponent(): JSX.Element {
...
-    }, [
-        table.statisticDataAbsolute,
-        setting.raw,
-        setting.stat,
-        isolateFilteredUrl,
-    ]);
+        if (isExport) {
+            fetchAndChooseData(setting.raw, setting.stat);
+        }
+    }, [isExport]);
```
GerberI, line 108: I think the effect and export state is not necessary here, you can just use:
```
handleExport = () => {
  setIsOpen(false);
  fetchAndChooseData(...);
}
```

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -88,13 +102,10 @@ export function HeaderExportContainerComponent(): JSX.Element {
     };
 
     useEffect(() => {
-        fetchAndChooseData(setting.raw, setting.stat);
-    }, [
-        table.statisticDataAbsolute,
-        setting.raw,
-        setting.stat,
-        isolateFilteredUrl,
-    ]);
```
GerberI, line 97: I think this was a huge performance bottleneck :)

GerberI (changes requested):
Commit: Only fetch data by clicking export

src/app/App.tsx
GerberI: If it is triggered if the page is loaded and not only after the user navigates through the app, yes.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -335,7 +330,7 @@ export function QueryPageContainerComponent(): JSX.Element {
     return (
         <LoadingOrErrorComponent
             status={{ isolateStatus, isolateCountStatus, filterStatus }}
-            dataIsSet={!_.isEmpty(data.ZNData)}
+            dataIsSet={!_.isEmpty(data.uniqueValues)}
```
GerberI, line 333: I think it would be better to derive the dataIsSet value from the different status states, so the loading state would not rely on one piece of data.

src/app/Shared/Context/DataContext.tsx
```diff
@@ -1,14 +1,11 @@
 import React, { useState, createContext } from "react";
-import { DbCollection } from "../Model/Client_Isolate.model";
 import { FilterInterface } from "../Model/Filter.model";
 
 interface DataInterface {
-    ZNData: DbCollection;
```
GerberI, line 6: I wonder if ZNData was used at all?

GerberI (approved):
Commit: Remove fetching of all isolates on QueryPage

src/app/Shared/Context/DataContext.tsx
MicheleK17: Only to get the length of the full dataset. Now it is retrieved over the counted isolates endpoint without filter.

src/app/Main/Header/Export/Dialog/Header-ExportDialog.component.tsx
MicheleK17: Done. Commit "Remove "isOpen" to control Dialog"

src/app/Main/Header/Header.component.tsx
MicheleK17: Commit: "Remove "display: none" situations "

GerberI (approved):
Commit: Remove "display: none" situations
Cool, thx

src/app/Main/Body/QueryPage/Drawer/Filter/Dialog/FilterSetting-Dialog.component.tsx
```diff
@@ -111,7 +111,6 @@ export function FilterSettingDialogComponent(props: {
     const cancelFilterToDisplayButton = t("FilterDialog.Cancel");
 
     return DialogComponent({
-        isOpen: props.isOpen,
```
GerberI, line 114: props.isOpen can be removed too

src/app/Main/Body/QueryPage/Drawer/Filter/Drawer-Filter.component.tsx
```diff
@@ -115,13 +117,15 @@ export function DrawerFilterComponent(props: {
...
-            <FilterSettingDialogComponent
-                isOpen={filterDialogIsOpen}
-                previousFiltersToDisplay={props.filterInfo.displayedFilters}
-                availableFilters={props.filterInfo.mainFilter}
-                onSubmitFiltersToDisplay={handleSubmitFiltersToDisplay}
-                onCancelFiltersToDisplay={handleCancelFiltersToDisplay}
-            />
+            {filterDialogIsOpen && (
+                <FilterSettingDialogComponent
+                    isOpen={filterDialogIsOpen}
```
GerberI, line 122: see other comment: can be removed

GerberI (changes requested):
Commit: Remove "isOpen" to control Dialog
thx

src/app/Main/Header/Header.component.tsx
MicheleK17: Done. Remove Style from Container.

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
MicheleK17: Done. Remove Style from Container

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
MicheleK17: The async export needs to be inside an Effect.

src/app/Main/Body/QueryPage/Drawer/Filter/Dialog/FilterSetting-Dialog.component.tsx
MicheleK17: Done. Remove Style from Container

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -115,15 +185,23 @@ export function HeaderExportContainerComponent(): JSX.Element {
         setIsOpen(true);
     };
     const handleClose = (): void => {
+        // eslint-disable-next-line unicorn/no-useless-undefined
+        setCsvDataString(undefined);
```
GerberI, line 150: you could set it to empty string

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -115,15 +185,23 @@ export function HeaderExportContainerComponent(): JSX.Element {
...
         setIsExport(true);
     };
 
-    const exportLabels = generateExportLabels(filter.mainFilter);
-    const buttonLabel: JSX.Element = ExportButtonLabelComponent(isOpen);
+    const reset = useCallback(() => {
+        // eslint-disable-next-line no-console
+        console.log("reset")
+        // eslint-disable-next-line unicorn/no-useless-undefined
+        setCsvDataString(undefined);
```
GerberI, line 160: same here

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -38,53 +42,74 @@ export function HeaderExportContainerComponent(): JSX.Element {
     const [setting, setSetting] = useState<ExportInterface>(defaultExport);
```
GerberI, line 28: only state really needed here is stat and raw flags, all the other things are only used by fetching function in a stateless manner

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -98,15 +123,60 @@ export function HeaderExportContainerComponent(): JSX.Element {
...
+        }
+        setSetting(newSettings);
+        setLoading(false);
+    };    
+
+    // https://stackoverflow.com/questions/64817556/loading-data-asynchronously-and-downloading-csv-data-with-a-click-is-one-step-be
 
     useEffect(() => {
         if (isExport) {
             fetchAndChooseData(setting.raw, setting.stat);
```
GerberI, line 137: i would set is export=false only here, then the logic becomes clearer. Something like:
```
setIsLoading(true);
await fetchAndChooseData(...)
setIsLoading(false);
setIsExport(false);
```

src/app/Main/Header/Export/Dialog/Header-ExportDialog.component.tsx
```diff
@@ -32,36 +32,44 @@ export function HeaderExportDialogComponent(props: {
...
+            <div>
+                {exportButtonLabel}
+                <CSVDownload
+                    data={props.dataString}
+                    filename={props.ZNFilename}
+                    target="_blank"
+                />
+            </div>
+        );
+        handleClearData()
```
GerberI, line 68: I really don't get this, if the dialog is called with a download link, it gets closed and the link is erased? How does this work?

GerberI (commented):
Commit: Implement async data Export
1. The general problem here is that you use two different methods for exporting (auto via zip and by clicking via link). Perhaps it would be more convenient to the user and more easier for the coding to always export a zip file.
2. I think it is not the best idea and not necessary to block the dialog with a spinner after clicking export. The dialog should just close and then a spinner could be displayed somewhere else (e.g. status bar or something) or not at all. It is normal web behavior to wait some time after downloading something but the app still being response during that time.

GerberI (commented):
Commit: Remove Style from Container
thx

GerberI (commented):
Commit: npm audit fix
I wonder if the package-lock is in sync with package.json
So we should discuss maintaining package.json with Dominic in the near future

PR #300, opened 2021-06-10
Disable selector during loading data

MicheleK17:
- reset row/column selector after leaving page

Ticket: #299
Change-Id: I8b2502584884b954afe94e1c9d7583cc059a36c1


src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -302,10 +304,14 @@ export function QueryPageContainerComponent(): JSX.Element {
             setTableContext(isolateCountGroups, nrOfSelectedIsolates);
             setNrOfSelectedIsol(nrOfSelectedIsolates);
         }
+        setTableIsLoading(false)
```
dptolle, line 309: if callApiService fails or throws an error is setTableLoading never set back to 'false'?

MicheleK17: No it is independant of the api response status.

dptolle (approved):

PR #313, opened 2021-06-30
Only show selected filters in parameter box

MicheleK17:


src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -20,16 +18,12 @@ export function QueryPageParameterContentComponent(props: {
...
     const {selectedFilter} = props
 
-    const displayFilter: Record<string, string[]> = _.cloneDeep(
-        selectedFilter
-    );
-    mainFilter.forEach((filterElement: string) => {
-        if (selectedFilter[filterElement].length === 0) {
-            displayFilter[filterElement] = [t("Filters.All")];
-        } else {
+    const displayFilter: Record<string, string[]> = {}
```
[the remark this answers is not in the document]
GerberI: Reminder: Try to use built in collections like map:
```
const displayedFilters = new Map<string, string[]>();
...
displayedFilters.set(filterElement, selectedFilter[filterElement]);
...
for(let element of displayedFilters.keys()) {
  ...
}
```
These are more safe to use and common in other languages too.

GerberI (approved):
Commit: Change Parameter layout&only show selected filters

GerberI (approved):
Commit: Rename Filter => Filters

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -34,14 +34,17 @@ export function QueryPageParameterContentComponent(props: {
...
         const elements: JSX.Element[] = [];
-        Object.keys(displayFilters).forEach((element): void => {
-            elements.push(
-                <ParameterContentListComponent
-                    key={`parameter_list_${element}`}
-                    element={element}
-                    listElements={displayFilters[element]}
-                />
-            );
+        Object.keys(displayedFilters).forEach((element): void => {
```
GerberI, line 37: replace with ``for(let element of displayedFilters.keys())``. Then you don't need the undefined check. (with Object.keys you iterate over the map object with all it's internal stuff, so only use Object.keys for plain java script objects.

GerberI (changes requested):
Commit: Rename and use Map
thx, the idea is to introduce you to more advanced and common techniques.

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
MicheleK17: "for ... of" loops are not allowed by the linter. "iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations. eslint-no-restricted-syntax"

GerberI: lol, can you please research this rule and try to explain what it means or what the actual impact is? Common sense is to prefer clean code over performance, except a part of code is a performance bottleneck.

MicheleK17: I haven't used it though, because I noticed that I can also get the information directly from the selectedFilters without creating a new Map. 
But from now on, I will pay more attention to whether I can use Map.

GerberI (commented):
Commit: Renaming & Simplifying
Good

src/app/Main/Body/QueryPage/Parameter/QueryPage-ParameterContent.component.tsx
GerberI: Funny i thought of that too, but didn't blamed it because i thought there might be a reason why the selectedFilters are checked against the mainFilters.
For further reading: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Keyed_collections

PR #315, opened 2021-07-05
Add axis labels

MicheleK17:
Ticket: #281
Change-Id: Id66421068b4658ae11bcad50f93a34bcf0c1be1d

### DIFF
--- a/src/app/Main/Body/QueryPage/Results/Charts/BarChartResults.component.tsx
+++ b/src/app/Main/Body/QueryPage/Results/Charts/BarChartResults.component.tsx
@@ -7,8 +7,8 @@ import { ApexChartData } from "./ApexChart.model";
 export function BarChartResultsComponent(props: {
     chartData: ApexChartData;
     getPngDownloadUriRef: MutableRefObject<(() => Promise<string>) | null>;
-    lableXaxis: string;
-    lableYaxis: string;
+    xAxisLabel: string;
+    yAxisLabel: string;
 }): JSX.Element {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const barChartRef = useRef<any>(null);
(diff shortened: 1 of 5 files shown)


src/app/Main/Body/QueryPage/Results/Charts/BarChartResults.component.tsx
```diff
@@ -7,6 +7,8 @@ import { ApexChartData } from "./ApexChart.model";
 export function BarChartResultsComponent(props: {
     chartData: ApexChartData;
     getPngDownloadUriRef: MutableRefObject<(() => Promise<string>) | null>;
+    lableXaxis: string;
```
GerberI, line 10: better name would be xAxisLabel, yAxisLabel. In general the more generic part comes first and specific part last. E.g. data => dataTable => dataTableHeader => dataTableHeaderLabel => dataTableHeaderLabelSize and so on.

src/app/Main/Body/QueryPage/Results/Charts/QueryPageContent-BarChart.component.tsx
```diff
@@ -64,6 +65,26 @@ function processingTableDataToApexData(
...
 }
 
+function generateAxisLabels(
+    t: TFunction,
+    rowName: string,
+    colName: string
+): [string, string] {
+    let lableXaxis = t("Results.TableHead");
+    let lableYaxis = "";
+    if (!_.isEmpty(rowName)) {
```
GerberI, line 75: I try to omit using external libraries whenever possible. So here ``if(rowName !== "")`` would do the same thing. But if you like using lodash you can keep it because it is a very common library.

GerberI (approved):
Good

src/app/Main/Body/QueryPage/Results/Charts/BarChartResults.component.tsx
MicheleK17: Done.

src/app/Main/Body/QueryPage/Results/Charts/QueryPageContent-BarChart.component.tsx
MicheleK17: True, in this case it is probably a bit too much to use Lodash.

GerberI (commented):
Commit: Rename & Lodash
thx, approved

PR #316, opened 2021-07-05
Change style

MicheleK17:
- sticky footer
    - scrolling behavior QueryPage
    - remove useless css

Ticket: #239
Change-Id: I29516e6e50cc69168a008978918456d3d2b35c68

### DIFF
--- a/src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
@@ -23,9 +23,11 @@ const useStyles = makeStyles(() =>
             width: "inherit",
             zIndex: 0,
             position: "relative",
+            direction: "rtl",
+            overflow: "auto",
         },
         drawerContainer: {
-            overflow: "auto",
+            direction: "ltr",
             padding: "1em",
             height: "100%",
         },
@@ -80,7 +82,9 @@ export function DrawerLayoutComponent(props: DrawerLayoutProps): JSX.Element {
     ): void => props.onFilterChange(selectedOption, keyName);
     const handleRemoveAllFilter = (): void => props.onFilterRemoveAll();
 
-    const handleSubmitFiltersToDisplay = (newFiltersToDisplay: string[]): void => {
+    const handleSubmitFiltersToDisplay = (
+        newFiltersToDisplay: string[]
+    ): void => {
         props.onSubmitFiltersToDisplay(newFiltersToDisplay);
     };
 
--- a/src/app/Main/Body/QueryPage/Parameter/ParameterContent-List.component.tsx
+++ b/src/app/Main/Body/QueryPage/Parameter/ParameterContent-List.component.tsx
@@ -20,6 +20,9 @@ const listStyle = css`
     padding: 0;
     margin: 0 4em 0 0;
 `;
+const listHeaderStyle = css`
+    position: relative;
+`;
 const parameterValue = css`
     margin-top: 0;
     margin-left: 2em;
@@ -86,6 +89,7 @@ export function ParameterContentListComponent(props: ListProps): JSX.Element {
                     component="div"
                     id="nested-list-subheader"
                     className={classes.nested}
+                    css={listHeaderStyle}
                 >
                     {label}
                 </ListSubheader>
--- a/src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
+++ b/src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
@@ -15,8 +15,6 @@ import { QueryPageContentBarChartResultsComponent } from "./Results/Charts/Query
 const contentStyle = css`
     width: 0;
     height: inherit;
-    margin-right: 1em;
-    margin-left: 1em;
     padding: 2em;
     flex: 1 1 0;
     hyphens: auto;
@@ -24,13 +22,8 @@ const contentStyle = css`
     overflow: auto;
 `;
 const contentBoxStyle = css`
-    max-width: 60em;
-    min-width: 20em;
+    max-width: 45em;
     margin: auto;
-    display: flex;
-    flex: 0 1 auto;
-    flex-direction: column;
-    hyphens: auto;
     box-sizing: border-box;
 `;
 const headingStyle = css`
@@ -73,7 +66,7 @@ export function QueryPageContentLayoutComponent(props: {
 
     return (
         <div css={contentStyle}>
-            <h1 css={headingStyle}>{t("Content.Title")}</h1>
+            <p css={headingStyle}>{t("Content.Title")}</p>
             <QueryPageDatabaseStatusIndicatorComponent />
             <div css={contentBoxStyle}>
                 {props.isFilter ? (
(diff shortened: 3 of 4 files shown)


src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
```diff
@@ -15,22 +15,15 @@ import { QueryPageContentBarChartResultsComponent } from "./Results/Charts/Query
 const contentStyle = css`
     width: 0;
     height: inherit;
```
GerberI, line 17: i think this can be removed: from the inherit mdn docs: 'For non-inherited properties, this specifies a behavior that typically makes relatively little sense and you may consider using initial instead, or unset on the all property.'

src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
```diff
@@ -15,22 +15,15 @@ import { QueryPageContentBarChartResultsComponent } from "./Results/Charts/Query
 const contentStyle = css`
     width: 0;
```
GerberI, line 16: see flex

src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
```diff
@@ -15,22 +15,15 @@ import { QueryPageContentBarChartResultsComponent } from "./Results/Charts/Query
 const contentStyle = css`
     width: 0;
     height: inherit;
-    margin-right: 1em;
-    margin-left: 1em;
     padding: 2em;
     flex: 1 1 0;
```
GerberI, line 19: I suggest to write a ticket to place all flex properties of flex items to the components where the flex box is defined. E.g. this flex should be in a div surrounding this component. The div should be placed in the QueryPage-Layout component. The reason is that the flex property here controls the behavior of the QueryPage-Layout. The QueryPageContent-Layout component should not depend on where it is used, so not depent on css properties in its parent component.

src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
```diff
@@ -15,22 +15,15 @@ import { QueryPageContentBarChartResultsComponent } from "./Results/Charts/Query
...
     hyphens: auto;
     box-sizing: border-box;
     overflow: auto;
 `;
 const contentBoxStyle = css`
-    max-width: 60em;
-    min-width: 20em;
+    max-width: 45em;
     margin: auto;
-    display: flex;
```
GerberI, line 30: Good, block elements are already layouted in a column, so using a flex is only necessary if you want to use wrapping or distribute the space between the block elements in a specific way.

src/app/Main/Main-Layout.component.tsx
```diff
@@ -12,29 +12,10 @@ const wrapperStyle = css`
     box-sizing: border-box;
 `;
 const bodyStyle = css`
-    width: 100%;
-    flex: 1 0 auto;
+    flex: 1;
```
GerberI, line 15: I recommended to use the long form `` 1 1 0``. E.g. i had to look in to the docs to check what the short form stands for.

GerberI (approved):
Good, you used the same technique as i used in the mibi-portal for fixing the footer.
For info: I had to set ``height: max-content`` for the flex items to make this work in apple browsers (safari, opera).

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
```diff
@@ -23,9 +23,11 @@ const useStyles = makeStyles(() =>
             width: "inherit",
             zIndex: 0,
             position: "relative",
+            direction: "rtl",
```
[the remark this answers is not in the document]
MicheleK17: I want the scrollbar to be on the left, so I need to display the content from right to left.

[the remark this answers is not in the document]
GerberI: Cool, nice trick. But in my opinion i would not mess with the scrollbar position, because scrollbars are always on the right.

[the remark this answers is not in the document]
MicheleK17: Its the scrollbar of the drawer. so it would be in the middle of the page if I don't put it on the left side. It is agreed with Katja and Dominic.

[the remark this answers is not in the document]
GerberI: ack

src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
MicheleK17: I don't understand this comment.

src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
MicheleK17: Done.

src/app/Main/Main-Layout.component.tsx
MicheleK17: Done.

src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
MicheleK17: Ticket is written.

src/app/Main/Body/QueryPage/QueryPageContent-Layout.component.tsx
GerberI: I think you used ``width: 0`` for the initial width so the content has the lowest flex priority and only can consume the remaining space in the flex row. So it also belongs to the QueryPage-Layout.

GerberI (approved):
Commit: Little Changes
thx

src/app/Main/Body/QueryPage/Parameter/ParameterContent-List.component.tsx
```diff
@@ -20,6 +20,9 @@ const listStyle = css`
     padding: 0;
     margin: 0 4em 0 0;
 `;
+const listHeaderStyle = css`
+    position: relative;
```
GerberI, line 24: Can you explain what the bug was? (it is to refresh my css knowledge)

GerberI (approved):
Commit: Fix scrolling bug in Parameter heading

GerberI:
All done, can be merged

src/app/Main/Body/QueryPage/Parameter/ParameterContent-List.component.tsx
MicheleK17: The headings of the parameters did not move when scrolling. They were visible until the accoridon disappeared completely from the visible field.  I hope the description is understandable.

GerberI: Sorry, i messed up my comment, i had another question: I wanted to know if the position is set to absolute by default on material-ui ListSubHeader components and if that's the cause.

PR #330, opened 2021-07-14
Add translation to table

MicheleK17:
Ticket: #301
Change-Id: I87c3e2ef744af558aa965e1838f1089421cb2334


src/app/Main/Body/QueryPage/Results/TableResults/TableContent-Header.component.tsx
```diff
@@ -66,14 +67,15 @@ export function TableContentHeaderComponent(props: {
...
-            } else {
-                headerTitle = `${t("Results.TableHeadRelative")}`;
-            }
+        let headerTitle = t(`FilterValues.${props.colAttribute}.${headerValue}`) ;
+        if (props.tableOption === "relative" && props.isCol) {
+            headerTitle = `${headerTitle} ${t("Results.Unit")}`;
+        } else if (props.tableOption === "relative" && !props.isCol) {
+            headerTitle = t("Results.TableHeadRelative");
+        } else if (props.tableOption === "absolute" && !props.isCol) {
+            headerTitle = t("Results.TableHead");
```
GerberI, line 76: Suggestion to make the code more understandable: The main difference here is if the header is a data header or not, so i would differentiate this first and then branch for the data type:
```
let headerTitle = '';
if(isCol) {
  headerTitle = t(...);
  if(relative) {
    headerTitle = headerTitle + t(...)
  }
else {
  if(relative) t(...)
  else if(absolute) t(...)
}
```

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-TableContainer.component.tsx
```diff
@@ -43,11 +44,13 @@ const useStyles = makeStyles({
  * @returns {JSX.Element} - table container component
  */
 export function TableContentTableContainerComponent(props: {
-    sumOptions: { showRowSum: boolean; showColSum: boolean };
-    isCol: boolean;
+    sumOptions: SumOptions;
     tableData: Record<string, string>[];
     displayOption: DisplayOptionType;
+    displayRowCol: DisplayRowCol;
```
GerberI, line 50: I do not understand what this means. Can you please explain (perhaps again) what the difference is if isRow or isCol is true or false? Has this something todo if the user already selected what to display or not?

src/app/Main/Body/QueryPage/Results/TableResults/TableResults.model.ts
```diff
@@ -1,4 +1,9 @@
 export interface SumOptions {
     showRowSum: boolean,
     showColSum: boolean
-  }
\ No newline at end of file
+  }
+
+export interface DisplayRowCol {
```
GerberI, line 6: As i commented, i do not really understand what that flags do. So there should be a better name for that interface (and the component's props) making things more clear.
In general an interface with more than one property should not be named with a verb but with a name describing the data the interface encapsulates. So `SumOptions` is a good name, `DisplayRowCol` not.
Also names for exported interfaces in your model files (names for your component props too) should be selected carefully, these are part of the code another developer would look at first to understand your code and the underlying structure of the data.

GerberI (commented):
Good.
Refactoring consideration: If you happen to refactor the table components i would move the translation specific parts to the tableContainer. So the view components of the table could be more stupid and just display the strings they get from the container without knowing about the actual data and it's structure.

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-TableContainer.component.tsx
MicheleK17: The frontend has two feature selectors. One for the attribute in the rows and one for the columns. The display works when both are selected, but also when only one is selected. So the question is "isCol? True : False" and "isRow? True : False".

GerberI: Thanks, i remember now, if only one axis is selected only the sum of the selected isolates is shown.

src/app/Main/Body/QueryPage/Results/TableResults/TableResults.model.ts
MicheleK17: What would you suggest?

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-Header.component.tsx
MicheleK17: Done. Change if-else

src/app/Main/Body/QueryPage/Results/TableResults/TableResults.model.ts
GerberI: Hehe, i would have given a suggestion if i had a good one.
So `DisplayOptions` would be to general, `DataConfiguration` is a little bit better. 
In general you should handle your table more like a multi-dimensional table with two dimensions (as we discussed also for the api), because the type of data in rows and cols is the same, so there are no rows representing entries and cols representing attributes of those entries. Your app data should be represented with such a structure and your table specific data should be data independent and just represent headers, rows, cols and the cells.
The clearest solution would be to represent the data configuration like:
```
interface IsolateDataFormat {
  rowAttribute: IsolateAttribute | null,
  colAttribute: IsolateAttribute | null
```
and change the attributes accordingly to the user input to the table options (data to show in row, col). And then derive the  table data from this in the results container: 
```
  if(dataFormat.rowAttribute !== null) {
    addRowHeaders(dataFormat.rowAttribute)
  } else if(...)
```
So in the end isCol and isRow are derived states that do not need to be modeled explicitly (but they can still be used as local variables).

src/app/Main/Body/QueryPage/Services/TableServices/calculateRelativeTableData.service.ts
```diff
@@ -11,17 +11,11 @@ function calculateRelative(part: number, total: number): string {
...
     objList: Record<string, string>[], 
     nrOfSelectedIsolates: number,
-    tableHeadRelative: string,
-    isRow: boolean,
 ): Record<string, string>[] {
     const statisticTableDataRel: Record<string, string>[] = [];
     objList.forEach((obj) => {
         const relativeRow: Record<string, string> = { name: obj.name };
-        if (isRow) {
-            relativeRow.name = obj.name
```
GerberI, line 21: Just for interest: Is obj.name === undefined or something if isRow === false?

GerberI (approved):
Commit: Change if-else
thx, also good you've separated the general row heading from the data calculation service and used the same pattern as for the cols. Next step would be to separate headers from the actual data at all (as stated in previous comment, to move more to the data as a multi-dimensional table direction and not modeling a header row and single row headers as in a conventional table).

GerberI:
Additional info just for fun: As for computer science, the data model is the most essential part, so if you implement an algorithm you first think about how the data is formed the algorithm is working on. It is essential to how fast, stable and complex an algorithm will be. So implementing good algorithm is more like finding good data models.

src/app/Main/Body/QueryPage/Results/TableResults/TableResults.model.ts
MicheleK17: Like this? Commit: "Check isCol?|isRow? local (no prop)"

src/app/Main/Body/QueryPage/Services/TableServices/calculateRelativeTableData.service.ts
MicheleK17: obj.name is never undefined. 
if isRow === flase, a general name,  like "Anzahl Isoalte", is shown.

GerberI (approved):
Commit: Check isCol?|isRow? local (no prop)
Nice. So i was not sure if removing isCol isRow was possible, but it was :)

src/app/Main/Body/QueryPage/Results/TableResults/TableResults.model.ts
GerberI: Yes, you did exactly what i thought of. Kind of overhead for just a renaming proposal but it makes the code better.
If you have the time (outside this ticket) you could dive into two things to make the row/col !== null checks more convenient:
First you could check if react proposes some data selection functionality. I think the reducer mechanism should have something like it. The idea here is that you store your app data via reducers and access it via selectors. The selectors are just a kind of function to get peaces of the data in the required form. Something like: `selectIsRowDataDefined(tableData)`. If not you can define your own getter functions beside the model definition like:
```
export interface TableData ...
export function getIsRowDataDefined(tableData: TableData): boolean {...}
```
Second, you could check if it is possible to use classes with react state. This would be the more object-oriented encapsulated way. Then you could define your tableData as a class with integrated accessors:
```
export class TableData {
  ...
  public get isRowDataDefined(): boolean { return this.row !== undefined }
}
```
So the idea is to find a way to handle derived data with react. For our angular projects we use a store mechanism with reducers and selectors.

MicheleK17: Make a new ticket for row/col !== null. Using classes with React state is not recommended.

GerberI (changes requested):
Commit: Rename TableContext to DataContext
Thx for clarifying things by renaming. Also cool that you removed the unnecessary context object.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -48,35 +47,36 @@ export function QueryPageContainerComponent(): JSX.Element {
     const [columnNameValues, setColumnNameValues] = useState<string[]>([]);
     const [totalNrOfIsol, setTotalNrOfIsol] = useState<number>(0);
     const [nrOfSelectedIsol, setNrOfSelectedIsol] = useState<number>(0);
-    const [tableIsLoading, setTableIsLoading] = useState<boolean>(false);
+    const [dataIsLoading, setTableIsLoading] = useState<boolean>(false);
```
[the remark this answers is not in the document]
MicheleK17: Done. Thx.

GerberI (approved):
Commit: Rename setTableIsLoading => setDataIsLoading

GerberI:
All done, can be merged

PR #332, opened 2021-07-15
Adapt closeMenuOnSelect prop of selector

MicheleK17:
Ticket: #305
Change-Id: I966b7ce773fe9c73d8fee00209048ee59d919578

### DIFF
--- a/src/app/Shared/Selector.component.tsx
+++ b/src/app/Shared/Selector.component.tsx
@@ -143,7 +143,7 @@ export function SelectorComponent(props: SelectorProps): JSX.Element {
             </InputLabel>
             <Select
                 css={selectAreaStyle}
-                closeMenuOnSelect={false}
+                closeMenuOnSelect={!props.isMulti}
                 isMulti={props.isMulti}
                 noOptionsMessage={() => props.noOptionLabel}
                 options={props.dropDownValuesObj}


GerberI (approved):
Lol, that was an easy one :)
So i can add a consideration regarding the tickets: I recommend to use more user focused ticket names. So someone looking through your commits is interested in what are the general changes for the user and not about the technical details.
So a good name for this ticket could be: `Close filter drop-down on selection` and then as an additional description something like: `Only non multi-select filter drop drop-down lists are closed.`

MicheleK17:
> 
> 
> Lol, that was an easy one :)
> So i can add a consideration regarding the tickets: I recommend to use more user focused ticket names. So someone looking through your commits is interested in what are the general changes for the user and not about the technical details.
> So a good name for this ticket could be: `Close filter drop-down on selection` and then as an additional description something like: `Only non multi-select filter drop drop-down lists are closed.`

The Name of the Ticket is written as a User Story ("As a User I would like to have the single select drop downs close after making a selection") - That is also what is displayed on the board. "Adapt closeMenuOnSelect prop of selector" is only the name of the Code Review. Thought it would make it easier for you to understand what I changed.

GerberI:
Ok, thanks, that makes sense.
But the review name is also used for the final commit as you can see on github commit history. So here you should somehow use the user story name.

PR #334, opened 2021-07-20
Add total sum to Table

MicheleK17:
Ticket: #287
Change-Id: Iffde29f4a66ff06c8711e14c2e8ba70a48cc9ad6

### DIFF
--- a/src/app/Main/Body/QueryPage/Results/TableResults/TableContent-RowWithColSums.component.tsx
+++ b/src/app/Main/Body/QueryPage/Results/TableResults/TableContent-RowWithColSums.component.tsx
@@ -2,7 +2,6 @@
 import { css, jsx, SerializedStyles } from "@emotion/core";
 import TableCell from "@material-ui/core/TableCell";
 import _ from "lodash";
-import { backgroundColor } from "../../../../../Shared/Style/Style-MainTheme.component";
 import {
     highlightedTableBorder,
     fixedCellSize,
@@ -19,12 +18,12 @@ const tableCellStyle = (isName: boolean): SerializedStyles => css`
     min-width: ${isName ? `${fixedCellSize}px` : "auto"};
 `;
 
-const emptyCellStyle = css`
+const totalSumStyle = css`
     border-right: none;
     border-bottom: none;
     border-top: ${highlightedTableBorder};
     border-left: ${highlightedTableBorder};
-    background-color: ${backgroundColor};
+    background-color: ${sumRowColBackgroundColor};
 `;
 
 function calculateColSum(
@@ -63,11 +62,15 @@ export function TableContentRowWithColSumComponent(props: {
         </TableCell>,
     ];
 
+    let totalColSum = 0
+
     props.headerValues.forEach((headerValue) => {
         let colSum: number | string = calculateColSum(
             props.tableData,
             headerValue
         );
+
+        totalColSum += colSum
 
         if (props.displayOption === "relative") {
             colSum = colSum.toFixed(2);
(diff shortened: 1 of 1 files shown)


MicheleK17:
Does it make sense to test the number in this component? 
The relative number must always be 100 and the absolute number must be the number of isolates. Otherwise throw an error. Or something like that?

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-RowWithColSums.component.tsx
```diff
@@ -63,12 +62,16 @@ export function TableContentRowWithColSumComponent(props: {
...
 
+    let totalColSum = 0
+
     props.headerValues.forEach((headerValue) => {
         let colSum: number | string = calculateColSum(
             props.tableData,
             headerValue
         );
 
+        totalColSum += colSum
```
GerberI, line 73: You use colSum as a container for number and string representation, so adding here is not very fine. I suggest to use an additional colSumString variable as you've done for the totalSum below.

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-RowWithColSums.component.tsx
```diff
@@ -63,12 +62,16 @@ export function TableContentRowWithColSumComponent(props: {
         </TableCell>,
     ];
 
+    let totalColSum = 0
```
GerberI, line 65: I wonder if this info also comes from the API? If so i would prefer the API given total sum.

GerberI (changes requested):

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-RowWithColSums.component.tsx
GerberI: Ok, you mentioned the total number of isolates in your question about testing. So i would use the API given number, then testing here becomes obsolete.

MicheleK17: But then the individual sums come from the frontend and the total sum from the API. shouldn't I test then even more that both numbers match?

GerberI: Ok, you are right, then you should leave it as it is. The table component does the summation and it is a feature of this component. Or perhaps later of the ResultsContainer and a proper summation service as the summation could be also relevant for the Chart.
As for the test, it would be one crossing concerns (api and table view). What you could do or want to test here is summing the isolates after getting them from the api and check if the total sum is the right one, here you could throw an exception. As for the summation in the table component, here you would rely on that component and a runtime check would be overhead, e.g. you would also not check the outputs of any library. (That's not true for applications maintaining nuclear power plants!).
Instead you would check the algorithm with unit and e2e tests before you commit new code:
Unit Test: You could check if the row and col sums the table component calculates are the same or match the 100%.
E2E Test: You could check if the displayed total sum in the table is the same as the displayed total sum in the results info header.
As for the 100% you could just write that value to the cell.

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-RowWithColSums.component.tsx
MicheleK17: Done.

src/app/Main/Body/QueryPage/Results/TableResults/TableContent-RowWithColSums.component.tsx
MicheleK17: Okay. So I made an extra Ticket for testing?

GerberI: You could make an epic Ticket for e.g. writing e2e-Tests and add that specific test to list in the description, to collect things the tests should cover.
In general i would start to write e2e tests, because they cover more parts of your code at ones. But the integration of them into the deployment environment is a job for Dominic.

GerberI (approved):
Commit: Add colSumString
Good, thx

GerberI:
All done, can be merged

PR #335, opened 2021-07-21
Change order of main filters

MicheleK17:
Ticket #253

Change-Id: Ie3874463a219b60b8dae746d71f95e1a4b1c32b8

### DIFF
--- a/src/app/Shared/Model/Client_Isolate.model.ts
+++ b/src/app/Shared/Model/Client_Isolate.model.ts
@@ -53,7 +53,7 @@ export type DbCollection = (Record<DbKeyWithStringValue, string> & {
     resistance: ResistantValue[];
 })[];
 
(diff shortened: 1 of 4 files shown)


src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -55,14 +55,14 @@ export type DbCollection = (Record<DbKeyWithStringValue, string> & {
 
 export const DbKeyCollection: DbKey[] = [
```
GerberI, line 56: If this array defines the filters in the app and not the keys in the DB, i suggest to rename it to something like:
MainFilterEntries or MainFilterList.

GerberI (approved):

GerberI (approved):
Commit: Rename
thx

GerberI:
can be merged

Issue #319, opened 2021-07-07
Chat overflows if Drawer is resized 


MicheleK17:
caused by "display: grid" to center chart

PR #337, opened 2021-07-26
Fix grow/shrink of chart

MicheleK17:
Ticket: #319
Change-Id: I26493898491bf4cedc46036a206bee2c4e579fff

### DIFF
--- a/src/app/Main/Body/QueryPage/Results/Charts/QueryPageContent-BarChart.component.tsx
+++ b/src/app/Main/Body/QueryPage/Results/Charts/QueryPageContent-BarChart.component.tsx
@@ -21,11 +21,12 @@ import { BarChartResultsComponent } from "./BarChartResults.component";
 
 const dataStyle = css`
     margin: 0 auto;
-    display: grid;
     box-sizing: border-box;
 `;
 const centerChartStyle = css`
     margin: auto;
+    overflow: auto;
+    flex: 1 1 0;
 `;
 const explanationTextStyle = css`
     text-align: center;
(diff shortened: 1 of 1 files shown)


src/app/Main/Body/QueryPage/Results/Charts/QueryPageContent-BarChart.component.tsx
```diff
@@ -21,11 +21,12 @@ import { BarChartResultsComponent } from "./BarChartResults.component";
 
 const dataStyle = css`
     margin: 0 auto;
-    display: grid;
     box-sizing: border-box;
 `;
 const centerChartStyle = css`
     margin: auto;
+    overflow: auto;
+    flex: 1 1 0;
```
GerberI, line 29: Remove: centerChartStyle is not set to `display: flex`, so the flex prop is useless here

src/app/Main/Body/QueryPage/Results/Charts/QueryPageContent-BarChart.component.tsx
```diff
@@ -21,11 +21,12 @@ import { BarChartResultsComponent } from "./BarChartResults.component";
 
 const dataStyle = css`
     margin: 0 auto;
-    display: grid;
```
GerberI, line 24: For info: The grid is not the real problem here. The problem is the `margin: auto` in centerChartStyle not working on grid-items. In general `margin: auto` is not the best way to center things with flex and grid, they have their own ways to handle centering. In the mibi-portal the only thing using margin: auto is the loading spinner, but it's because the animation needs it to work properly.

GerberI (changes requested):
Commit: Fix grow/shrink of chart

src/app/Main/Body/QueryPage/Results/Charts/QueryPageContent-BarChart.component.tsx
MicheleK17: Done.

GerberI (approved):

GerberI (approved):

GerberI:
can be merged

PR #338, opened 2021-07-26
Remove info about row/col from isolate url

MicheleK17:
Ticket: #275
Change-Id: I5188b4749a16c016b795377bf6de2bf5e1935e24

### DIFF
--- a/src/app/Main/Header/Export/Header-ExportContainer.component.tsx
+++ b/src/app/Main/Header/Export/Header-ExportContainer.component.tsx
@@ -61,7 +61,10 @@ export function HeaderExportContainerComponent(): JSX.Element {
         setting.tableAttributeNames.column = undefined
     }
 
-    const isolateFilteredUrl: string = ISOLATE_URL + history.location.search;
+    const filterSettingsPath = history.location.search.replace(/&row.+/g, "").replace(/&column.+/g, "")
+
+    const isolateFilteredUrl: string = ISOLATE_URL + filterSettingsPath;
+
 
     const buttonLabel: JSX.Element = ExportButtonLabelComponent(isOpen);
 
--- a/src/app/Main/Header/Export/Header-ExportContainer.component.tsx
+++ b/src/app/Main/Header/Export/Header-ExportContainer.component.tsx
@@ -64,7 +64,8 @@ export function HeaderExportContainerComponent(): JSX.Element {
     const urlParams = new URLSearchParams(history.location.search);
     urlParams.delete("row");
     urlParams.delete("column");
-    const isolateFilteredUrl = `${ISOLATE_URL}/${urlParams}`;
+    const urlParamsString = urlParams.toString()
+    const isolateFilteredUrl = `${ISOLATE_URL}/${urlParamsString}`;
     const buttonLabel: JSX.Element = ExportButtonLabelComponent(isOpen);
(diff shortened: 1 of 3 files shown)


src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -61,7 +61,10 @@ export function HeaderExportContainerComponent(): JSX.Element {
         setting.tableAttributeNames.column = undefined
     }
 
-    const isolateFilteredUrl: string = ISOLATE_URL + history.location.search;
+    const filterSettingsPath = history.location.search.replace(/&row.+/g, "").replace(/&column.+/g, "")
```
GerberI, line 64: As stated in the grooming, just deleting here is not the best practice. E.g. if you happen to add another client side param you would have to delete it here to. Better is to pick the things you need here. You could try to use: ``history.location.state.params`` to get the params for your isolates query.

GerberI (changes requested):

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
MicheleK17: `history.location.state.params` is no valid command. Of cause I could extract all parameter of the url (which would include row and col as well). But then I would have to extract the filter parameters and reassemble them into a string. That seems like an overkill and I think I'll leave it as it is. At the moment, no further parameters for the URL are planned.

GerberI: you can try the hook `useLocation` and use `location.state.params` and check if this works. If not you can still use url parsing: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL to get the params.
If you want to leave it as it is, then you should at least ensure that only the desired part of the url is removed. Your regexp patterns just remove all things after `&row`. But url params can have any order and a user can change it and the client should still process it as a valid request. So even here you should use a the recommended utilities to do the work for you.

GerberI (changes requested):

src/app/Main/Header/Export/Header-ExportContainer.component.tsx
```diff
@@ -58,14 +58,13 @@ export function HeaderExportContainerComponent(): JSX.Element {
...
 
-    const filterSettingsPath = history.location.search.replace(/&row.+/g, "").replace(/&column.+/g, "")
-
-    const isolateFilteredUrl: string = ISOLATE_URL + filterSettingsPath;
-
-
+    const urlParams = new URLSearchParams(history.location.search);
+    urlParams.delete("row");
+    urlParams.delete("column");
+    const isolateFilteredUrl = `${ISOLATE_URL}/${urlParams}`;
```
GerberI, line 67: In general, i try to be as explicit as possible, so i would use: `urlParams.toString()` to make clear that some conversion from a params object to string happens here. Most developers wouldn't use that because in most languages toString() is defined on object (and therefore present on all classes) and called by default if used in a string expression.
So another reason i prefer the explicit way is to remind me of implementing toString() on my own classes :)

GerberI (approved):
Commit: Remove col and row parameters from url safely
Good, thx.
I didn't know that there is also an URLSearchParams utility class.

PR #354, opened 2021-09-28
Implement SubFilter

MicheleK17:
Ticket: #351
Change-Id: If4b1250a4ed01da1af3c8b5d26c64d1c38692fd2

### DIFF
--- a/src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
@@ -4,6 +4,7 @@ import Drawer from "@material-ui/core/Drawer";
 import { DrawerFilterComponent } from "./Filter/Drawer-Filter.component";
 import { DrawerDisplayedFeaturesComponent } from "./Displayed_Features/Drawer-DisplFeatures.component";
 import {
+    ClientSingleFilterConfig,
     FilterInterface,
     FilterType,
 } from "../../../../Shared/Model/Filter.model";
@@ -43,6 +44,7 @@ export interface DrawerLayoutProps {
     drawerWidth: number;
     dataUniqueValues: FilterInterface;
     filterInfo: FilterContextInterface;
+    subFilters: ClientSingleFilterConfig[];
     columnFeature: string;
     rowFeature: string;
     onDisplFeaturesChange: (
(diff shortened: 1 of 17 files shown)


src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
```diff
@@ -43,6 +44,7 @@ export interface DrawerLayoutProps {
     drawerWidth: number;
     dataUniqueValues: FilterInterface;
     filterInfo: FilterContextInterface;
+    subFilters: ClientSingleFilterConfig[];
```
dptolle, line 47: Not sure if it wouldn't make more sense to have the subfilters nested within filterInfo somehow.  After all subfilters by themselves can't really exist and only make sense in the context of main filters,  This would require a heavy rewrite of the FilterContextInterface I guess.

src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
```diff
@@ -31,6 +32,7 @@ export function QueryPageLayoutComponent(props: {
     };
     dataUniqueValues: FilterInterface;
     filterInfo: FilterContextInterface;
+    subFilters: ClientSingleFilterConfig[];
```
dptolle, line 35: Ask me at some stage about Interface pollution.

dptolle (approved):
Looks fine, but please take note of my comments.

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
MicheleK17: Yes, I get the point. That was my first thought too. But the information "Which filters are subfilters?" is only fetched once from the server when the component is loaded. But every change of the filters/filterInfo (which then also include the subfilters) generates a new call to the Isolate API. React doesn't like to store these two things in one hook (or I haven't found the right way yet).

MicheleK17: Another solution would be to hardcode the subfilters in the frontend, then I wouldn't need the subfilter-info from the api (parent/target) at all. Hard-coded stuff can also be saved in the filter hook. And since the translation takes place in the frontend, they are already hard-coded anyway.

PR #359, opened 2021-09-30
Change chart type to stacked if data is relative

MicheleK17:
Ticket: #342
Change-Id: I3c30a7cbbc6e1ef6a9a9e5b049786f6dd91bc183

### DIFF
--- a/src/app/Main/Body/QueryPage/Results/Charts/BarChartResults.component.tsx
+++ b/src/app/Main/Body/QueryPage/Results/Charts/BarChartResults.component.tsx
@@ -10,6 +10,7 @@ export function BarChartResultsComponent(props: {
     xAxisLabel: string;
     yAxisLabel: string;
     xAxisMax?: number;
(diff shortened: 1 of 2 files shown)


src/app/Main/Body/QueryPage/Results/Charts/BarChartResults.component.tsx
```diff
@@ -10,6 +10,7 @@ export function BarChartResultsComponent(props: {
     xAxisLabel: string;
     yAxisLabel: string;
     xAxisMax?: number;
+    isStackedBarChart: boolean;
```
GerberI, line 13: Consider using something else then `is` here. The variable is used as a setter from the outside, not a getter. So something like: `useStackedMode` or `displayAsStacked` or just `stacked` would be better.

GerberI (approved):

GerberI (approved):
thx

GerberI:
can be merged

Issue #346, opened 2021-08-05
Selecting another filter value without finishing loading the previous one will generate a warning from React and exit the page. 


MicheleK17:
- Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

PR #364, opened 2021-10-13
Disable selector while data is loading

MicheleK17:
Ticket: #346
Change-Id: I880bbe2bbd661c61c337cdcd956bf6bdcad858d2


GerberI (approved):
Commit: Disable selector while data is loading
I wonder if there was a loading spinner and if that one could be used instead of disabling the different components.

Issue #367, opened 2021-10-22
As a User I want my Filter value set reduced based on my previously chosen values


dptolle:
@dptolle commented on [Fri Oct 22 2021](https://github.com/SiLeBAT/zoonotify-server/issues/120)

PR #369, opened 2021-11-02
Fetch conditional filter values if new filter value is set

MicheleK17:
Ticket: #367
Change-Id: I6cf1c89c94580cbe3eeb6058a80790153c201bc5

### DIFF
--- a/src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
+++ b/src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
@@ -43,7 +43,10 @@ import { IsolateCountedDTO } from "../../../Shared/Model/Api_Isolate.model";
 import { FilterConfigDTO } from "../../../Shared/Model/Api_Filter.model";
 import { getCurrentDate } from "../../../Core/getCurrentDate.service";
 import { adaptFilterFromApiService } from "./Services/adaptFilterFromAPI.service";
-import { calculateRowColSumsAbsolute, calculateRowColSumsRelative } from "./Services/TableServices/calculateRowColSums.service";
+import {
+    calculateRowColSumsAbsolute,
+    calculateRowColSumsRelative,
+} from "./Services/TableServices/calculateRowColSums.service";
 
 export function QueryPageContainerComponent(): JSX.Element {
     const [isolateStatus, setIsolateStatus] = useState<number>();
@@ -52,7 +55,9 @@ export function QueryPageContainerComponent(): JSX.Element {
     const [columnNameValues, setColumnNameValues] = useState<string[]>([]);
     const [totalNrOfIsol, setTotalNrOfIsol] = useState<number>(0);
     const [nrOfSelectedIsol, setNrOfSelectedIsol] = useState<number>(0);
-    const [dataIsLoading, setDataIsLoading] = useState<boolean>(false);
+    const [dataIsMounted, setDataIsMounted] = useState<boolean>(false);
+    const [dataIsLoading, setDataIsLoading] = useState<boolean>(true);
+    const [conditionalValuesAreLoading, setConditionalValuesAreLoading] = useState<boolean>(true);
     const [uniqueDataValues, setUniqueDataValues] = useState<FilterInterface>(
         {}
     );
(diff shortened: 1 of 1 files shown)


src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -248,9 +254,52 @@ export function QueryPageContainerComponent(): JSX.Element {
...
+                        );
+                        newUniqueValues[mainFilter] =
+                            uniqueValuesObject[mainFilter];
+                    }
+                }
+            }
+        );
+
+        await Promise.all(getConditionalFilterValues);
+        setUniqueDataValues(newUniqueValues);
```
GerberI, line 291: For convenience and to prevent later bugs i would not rely on the error page displayed if an api call fails. Try to keep the data always in some kind of consistent way.
Solution 1: Keep the old filters and set the filters only if status = 200 (this would still be kind of inconsistent)
Solution 2: Assign and empty object if an error occurs so that no data is available
Solution 3 (possible the best one): Keep the data with all filter values from loading the page in a separate state variable and assign it to the actual values if a call fails.

GerberI (changes requested):
I think i have found a problem with the only-update-empty-filter solution.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -248,9 +254,52 @@ export function QueryPageContainerComponent(): JSX.Element {
...
     };
 
+    const fetchNewFilterValues = async (): Promise<void> => {
+        setConditionalValuesAreLoading(true);
+        const newUniqueValues = _.cloneDeep(uniqueDataValues);
+        const filterStatusList: number[] = [];
+        const getConditionalFilterValues = filter.displayedFilters.map(
+            async (mainFilter) => {
+                const selectedFilter = filter.selectedFilter[mainFilter];
+                if (_.isEmpty(selectedFilter)) {
```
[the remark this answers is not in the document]
MicheleK17: Yes, you are right. 
I'll have to discuss it with Dominic. 
Because when I fetch the conditional values for a selected filter, I only get back the selected filter value itself. 
For your example: 

- organism A: year 2017
- organism B: year 2018
- organism C: year 2017

URL: /filterconfig/organism/organism=organismA
API returns only "OrganismA" as filter value and not other organisms (Organism C) that would be possible for 2017.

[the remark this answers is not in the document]
MicheleK17: Yes, you'r right. 
It's done.

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: Done. (Solution 3)

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -263,41 +270,48 @@ export function QueryPageContainerComponent(): JSX.Element {
...
         const getConditionalFilterValues = filter.displayedFilters.map(
-            async (mainFilter) => {
-                const selectedFilter = filter.selectedFilter[mainFilter];
-                if (_.isEmpty(selectedFilter)) {
-                    const conditionalFilterUrl = `${FILTER_URL}/${mainFilter}/${history.location.search}`;
+            async (displayedFilter) => {
+                const urlParams = new URLSearchParams(history.location.search);
+                urlParams.delete("row");
+                urlParams.delete("column");
+                urlParams.delete(displayedFilter);
```
GerberI, line 277: This is for fixing the api issue you mentioned in your deleted comment?

GerberI (approved):
Commit: Load conditional Values for all filters & set initialUniqueValues
Good, thx

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
MicheleK17: Yes

Issue #139, opened 2020-10-29
Update to React v17


dptolle:
NB: Talk to Isaak beforehand

PR #387, opened 2021-11-23
Update to React 17

dptolle:
Update Material UI
	Replace obsolete i18next-xhr-backend
	Update react-dom
	Replace awesome-typescript-loader

Change-Id: I6b68cfc40ded79b5da9f1dbc6a3c41f933453bf2
Ticket: #139


src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
```diff
@@ -50,23 +25,40 @@ export function DrawerLayoutComponent(props: {
...
     return (
         <Drawer
-            className={classes.drawer}
-            variant="persistent"
-            anchor="left"
-            open={props.isOpen}
-            classes={{
-                paper: classes.drawerPaper,
-            }}
+          sx={{
```
MicheleK17, line 32: Why is the style now defined in the component and no longer with the help of "makeStyles"? In other components makeStyles is still included (inconsistent?).

MicheleK17 (approved):
Just one question for better understanding. Can be merged.

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
dptolle: Yeah - I had issues with CSS Specificity & not the right styles taking affect.  I've pushed the version with MakeStyles - we could commit that but the styles need to be adjusted again.  What do you think?

MicheleK17 (approved):

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
MicheleK17: Difficult for me to say because I don't see the frontend. But yes we can commit the update and in an extra ticket adjust the style where it is incorrect.

PR #390, opened 2021-11-25
Change drawer position to relative

MicheleK17:
Ticket: #388
Change-Id: If90433757336eef261c783a496c4fb96dd878227

### DIFF
--- a/src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
+++ b/src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
@@ -6,39 +6,39 @@ import makeStyles from "@mui/styles/makeStyles";
 import { primaryColor } from "../../../../Shared/Style/Style-MainTheme.component";
 
 const filterHeadingStyle = css`
-  margin-top: 0.5em;
-  margin-bottom: 0.5em;
-  font-weight: bold;
-  font-size: 1.5rem;
-  line-height: 2rem;
-  color: ${primaryColor};
+    margin-top: 0.5em;
+    margin-bottom: 0.5em;
+    font-weight: bold;
+    font-size: 1.5rem;
+    line-height: 2rem;
+    color: ${primaryColor};
 `;
 
 const ss = createStyles<
-  string,
-  {
-    drawerWidth: string;
-  }
+    string,
+    {
+        drawerWidth: string;
+    }
 >({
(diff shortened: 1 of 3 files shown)


src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
```diff
@@ -49,32 +49,38 @@ const useStyles = makeStyles(() => ss);
...
+    return (
+        <Drawer
+            className={classes.drawer}
+            variant="persistent"
+            anchor="left"
+            open={props.isOpen}
+            classes={{
+                paper: classes.drawerPaper,
+            }}
+            sx={{
```
GerberI, line 72: I don't understand this, looks like a hack. The cause of the problem might be elsewhere.
But for a fix it is ok.

src/app/Shared/ClearSelectorButton.component.tsx
```diff
@@ -47,18 +47,21 @@ export function ClearSelectorComponent(props: {
...
-                    <CancelIcon
-                        css={css`
-                            height: 20px;
-                            width: 20px;
-                        `}
-                    />
-                </IconButton>
+                <span>
+                    <IconButton
+                        css={iconButtonStyle}
```
GerberI, line 52: iconButtonStyle: display: inline-block or so could have worked too.

GerberI (approved):

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
MicheleK17: No this is not a hack. This is new by the update to material ui 5. In the long run the whole css will be changed to it.

src/app/Shared/ClearSelectorButton.component.tsx
MicheleK17: This is the official solution that Material Ui recommends.

GerberI: 👍

src/app/Main/Body/QueryPage/Drawer/Drawer-Layout.component.tsx
GerberI: And what does sx mean?
I think they want to make their components more customizable. Or what is the reason behind it?

PR #393, opened 2021-11-26
Put the shadow of the header in the foreground.

MicheleK17:
Ticket: #336
Change-Id: I7d70d41fa3caf435520dbc5854234c9829c119e8

### DIFF
--- a/src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
+++ b/src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
@@ -17,10 +17,10 @@ const subHeaderStyle = css`
     background-color: ${bfrPrimaryPalette[300]};
     box-sizing: border-box;
     box-shadow: 0 8px 6px -6px grey;
-    z-index: 10;
+    z-index: 1;
 `
-
 const queryPageStyle = css`
+    z-index: 0;
     display: flex;
     flex: 1 1 0;
     overflow: auto;
--- a/src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
+++ b/src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
@@ -17,6 +17,7 @@ const subHeaderStyle = css`
     background-color: ${bfrPrimaryPalette[300]};
     box-sizing: border-box;
     box-shadow: 0 8px 6px -6px grey;
+    z-index: 10;
 `
 
 const queryPageStyle = css`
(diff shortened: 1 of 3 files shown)


GerberI (changes requested):
Commit: Put the shadow of the header in the foreground.
Reject: Please don't use global z-indices. Use stacking context instead: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context
So in your case you should add a z-index: 0 to parents and ensure they build an own stacking context.

src/app/Main/Header/Header.component.tsx
```diff
@@ -16,6 +16,7 @@ const headerStyle = css`
     width: 100%;
     height: 100%;
     display: flex;
+    z-index: 0;
```
GerberI, line 19: still global, please remove

src/app/Main/Header/Header.component.tsx
```diff
@@ -28,7 +29,7 @@ const mainHeaderStyle = (open: boolean): SerializedStyles => css`
     justify-content: space-between;
     box-sizing: border-box;
     box-shadow: ${open ? "0 8px 6px -6px grey" : "none"};
-    z-index: 10;
+    z-index: 1;
```
GerberI, line 32: still global please remove.
The header should not be responsible of it's z position. It is the responsibility of the parent component (some layout?) to arrange the components. See also comment in QueryPageLayout how to do that.

src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
```diff
@@ -17,10 +17,10 @@ const subHeaderStyle = css`
     background-color: ${bfrPrimaryPalette[300]};
     box-sizing: border-box;
     box-shadow: 0 8px 6px -6px grey;
-    z-index: 10;
+    z-index: 1;
```
GerberI, line 20: Nope
This component should go like:
mainStyle is the parent and should establish a new stacking context. As documented use:
mainStyle: position: relative (also the default but it must be set to establish the stacking), z-index: 0
subHeaderStyle and queryPageStyle are your components you want to arrange in the stacking-context, so:
subHeaderStyle: z-index: 1
queryPageStyle: z-index: 0

src/app/Main/Body/QueryPage/QueryPage-Layout.component.tsx
```diff
@@ -17,10 +17,10 @@ const subHeaderStyle = css`
...
 `
-
 const queryPageStyle = css`
+    z-index: 0;
```
GerberI, line 23: Nope

GerberI (changes requested):
Commit: Use stacking context
Not correct.
A while ago i thought about to make it mandatory for developers to add a position: relative/absolute and z-index: 0 to each components main html element. This would make it impossible to mess up the z-indices.

MicheleK17:
Resolved on new branch "story/336_new"

PR #394, opened 2021-11-26
Remove dependency on legacy styling system

dptolle:
Ticket: #389


dptolle:
@MicheleK17 Some of the styles break - especially the drawer transition.  These should be fixed in a new Ticket

MicheleK17 (approved):

PR #398, opened 2021-12-01
Display all possible SubFilters

MicheleK17:
- adapt to new server information
    - add translation
    - adapt parameter box

Ticket: #397
Change-Id: I31f32930aa43ffba6f56ed0e47c4ef2d0c990ab5


src/app/Main/Body/QueryPage/Drawer/Filter/generateSelectorObject.service.ts
```diff
@@ -1,12 +1,23 @@
...
 export function generateSelectorObject(
     filterAttribute: string,
     selectorArray: string[],
-    t?: TFunction
+    t?: TFunction, 
+    isSubFilter?: boolean,
 ): { value: string; label: string }[] {
     return selectorArray.map((filterValue) => {
         if (t !== undefined) {
+            if (isSubFilter) {
```
GerberI, line 12: Consider a little refactoring here: 
```
let filterLabel = filterValue;
if(t !== undefined) {
  if(isSubFilter) {
    filterLabel = t(...)
  } else [
    filterLabel = t(...)
  }
}
return { value: filterValue, label: filterLabel }
```

src/app/Shared/Model/Api_Filter.model.ts
```diff
@@ -3,7 +3,7 @@ export interface FilterConfigDTO {
         id: string;
         name: string;
         parent: string;
-        target: string;
+        trigger: string;
```
GerberI, line 6: I would not say this is the better name...

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/ParameterContent-List.component.tsx
```diff
@@ -1,83 +1,22 @@
...
-                    css={parameterValue}
-                />
-            </ListItem>
-        );
-    });
-    return elements;
-}
 
 export interface ListProps {
     parameterLabel: FilterType;
```
GerberI, line 15: Reject: This is of type string now.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -27,11 +29,26 @@ export function QueryPageParameterContentComponent(props: {
         const elements: JSX.Element[] = [];
         Object.keys(selectedFilters).forEach((filterElement) => {
             if (selectedFilters[filterElement].length !== 0) {
+                let isSubFilter = false;
```
GerberI: As stated in another comment: Messing with the sub filters is clearly a part of a container. This component is just for displaying parameters and not for putting the data together it should display.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/ParameterContent-List.component.tsx
```diff
@@ -1,83 +1,22 @@
...
-    parameterList: string[];
+    parameterValuesList: JSX.Element[];
```
GerberI, line 16: Reject: 
First this is a layout now and you should name it accordingly.
Second: You have splitted the ``<List>`` specific elements now. You now assume that the caller of this component know that he should use ``<ListItems>`` as input components, so the encapsulation principle is broken.
Solutions:
1. If you want to behave this like a layout keep all the ``<List>`` specific elements in this components and adapt the input from the parent component accordingly.
2. Reverse your changes and let this component be a view again and use only the data this component should display as input (translated or not).

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/createParameterList.service.tsx
```diff
@@ -0,0 +1,64 @@
...
+`;
+const parameterValue = css`
+    margin-top: 0;
+    margin-left: 2em;
+    span {
+        letter-spacing: 0;
+    }
+`;
+
+export function createParameterListService(
```
GerberI: Reject: Please do not use component factories like this service. This is not necessary in your app and blurs the distinction between components and services. Styles and html code should be only used in conventional react components. So move this back to a component.
You could use a mapper services for extracting the translation stuff and create the data required by the parameterListComponent.

src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -65,7 +65,20 @@ export const MainFilterList: DbKey[] = [
     "federalState",
 ];
 
-export const subFiltersList: string[] = ["serovar", "species", "serotyp"];
+export const subFiltersList: string[] = [
```
GerberI, line 68: Are these a mix of matrix and microorganisms sub filters? If yes i would split them into two structures to keep them separate. You can use ``allSubFilters = [...matrixSubFilters, ...microSubFilters]`` if you need the merged list.

GerberI (changes requested):
Commit: Display all possible SubFilters
Quite some changes necessary here.

GerberI (approved):
Commit: Change parameter order & translate SubFilterTrigger
Approved when ignoring the rejects of the previous commit.

src/app/Shared/Model/Api_Filter.model.ts
MicheleK17: Dominic and I think this is better because it triggers the corresponding subfilter.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/QueryPage-ParameterContent.component.tsx
MicheleK17: This is part of the container refactoring ticket.

src/app/Main/Body/QueryPage/Drawer/Filter/Filter-SelectorList.component.tsx
```diff
@@ -51,25 +51,27 @@ export function FilterSelectorListComponent(
             );
             subFilters.forEach((subFilter) => {
                 if (filter === subFilter.parent) {
-                    const subFilterTarget = subFilter.target;
+                    const subFilterTrigger = subFilter.trigger;
                     if (
-                        subFilterTarget !== undefined &&
-                        _.includes(
+                        subFilterTrigger !== undefined &&
+                        (_.includes(
```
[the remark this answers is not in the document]
MicheleK17: This is part of the container refactoring ticket.

src/app/Shared/Model/Api_Filter.model.ts
GerberI: I think we had this discussion sometimes ago and i dismissed trigger as a name and proposed target. Trigger is kind of reversed for event like entities. So e.g. an error message banner could be triggered by an api failure. But it is more often used to describe triggering a chain of state transitions.
So the main problem with the name i have is, that it has behavioral character and gives the intention, as you stated, that the filterConfig can be triggered. But it is a simple data object holding data for describing the filter and should not wire in behavior. So what if the front-end does show all possible filters (as in the beginning), then the name would make no sense.
As also stated sometimes ago a more simpler approach would be to inverse the logic:
```
FilterDefinition: {
  value: string,
  subFilter?: Filter
}
Filter: {
  name: string,
  definitions: FilterDefinition[]
}  
```
(FilterDefinition is a bad name too :) ) So no parent, target or trigger.
This comment is just for give you some thinking :) As for the API is private and i think you all know what is meant by trigger the name might be fine.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/ParameterContent-List.component.tsx
MicheleK17: Done.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/createParameterList.service.tsx
MicheleK17: Done. But not sure if I solved it exactly as you imagined.

src/app/Main/Body/QueryPage/Drawer/Filter/generateSelectorObject.service.ts
MicheleK17: Done.

src/app/Main/Body/QueryPage/Drawer/Filter/generateSelectorObject.service.ts
```diff
@@ -1,12 +1,23 @@
 import { TFunction } from "i18next";
+import { replaceAll } from  "../../../../../Core/replaceAll.service";
 
 export function generateSelectorObject(
     filterAttribute: string,
     selectorArray: string[],
-    t?: TFunction
+    t?: TFunction, 
+    isSubFilter?: boolean,
```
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Shared/Model/Api_Filter.model.ts
MicheleK17: But the front end only shows the subfilter selector if the "trigger" was previously selected in a MainFilter selector. So it actually changes the behavior of the frontend.

GerberI: Yes, actually it does. I thought of another front-end or changing the front-end so it always shows all filters. Then it would make no sense. So, the API should be independent of the front-end implementation details and a REST API should only describe data and not behavior

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -33,18 +84,33 @@ function createParameterList(
...
+            selectedFilters[filterElement].forEach((parameter) => {
+                const parameterListItem = createParameterName(
+                    filterElement,
+                    parameter,
+                    t,
+                    isSubFilter
+                );
+                listItems.push(
+                    <ListItem
+                        key={`listItem-${parameterListItem.key}`}
```
GerberI, line 97: Is it intentional that `keyName = parameter.replace(" ", "_");` is not used as key here anymore?

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -2,22 +2,73 @@
...
+): { parameterName: string | JSX.Element; key: string } {
+    const key = parameter.replace(".", "");
+    let parameterName: string | JSX.Element = "";
+    if (parameterAttribute === "microorganism") {
+        const translateRootString = `FilterValues.formattedMicroorganisms.${key}`;
+        const prefix = t(`${translateRootString}.prefix`);
+        const name = t(`${translateRootString}.name`);
+        const italicName = t(`${translateRootString}.italicName`);
+        const suffix = t(`${translateRootString}.suffix`);
+        parameterName = getMicroorganismLabelService({
```
GerberI, line 52: So this is also a component-service mix, but that's another story...

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/QueryPage-ParameterContent.component.tsx
```diff
@@ -33,18 +84,33 @@ function createParameterList(
...
+            const listItems: JSX.Element[] = [];
+            selectedFilters[filterElement].forEach((parameter) => {
+                const parameterListItem = createParameterName(
+                    filterElement,
+                    parameter,
+                    t,
+                    isSubFilter
+                );
+                listItems.push(
+                    <ListItem
```
GerberI, line 96: Reject: ListItems and the corresponding List are still splitted over two components. You could just use the parameterName components as input for the ParameterListLayout and compose the List there.

GerberI (changes requested):
Commit: Rename & little Refactoring
thx for the little changes. The list issue is still only partly done.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/ParameterContent-List.component.tsx
GerberI: Solution 1 only party done, see comment in corresponding commit.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/createParameterList.service.tsx
GerberI: Not exactly, but then you have to refactor the microorgaLabelService too (i missed that). But the reject is done.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/QueryPage-ParameterContent.component.tsx
MicheleK17: ok.

src/app/Shared/Model/Client_Isolate.model.ts
MicheleK17: Done.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/QueryPage-ParameterContent.component.tsx
MicheleK17: Yes, because I hadn't edited that part yet. Now it is done.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/ParameterList-Layout.component.tsx
```diff
@@ -9,10 +9,24 @@ const listStyle = css`
...
+    span {
+        letter-spacing: 0;
+    }
+`;
 
 export interface ListProps {
     parameterLabel: string;
-    parameterValuesList: JSX.Element[];
+    parameterNames: {
+        parameterName: string | JSX.Element;
```
GerberI, line 27: As discussed sometimes ago: In general it would be better to pass data here only and let this component create the actual JSX.Elements for the names. So the involved services would not depend on react and just generate the necessary data. Then this component would be a clean view again (now it is some mix of a view and a layout).
But this would be something for a separat ticket.

GerberI (approved):
Commit: Refactoring and adapt matrixDetail
Thx, now i am happy :) For me it was a big code smell with the html list structure smeared over different components. It broke the encapsulation principle and would have made it harder to replace the list with something else.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/ParameterContent-List.component.tsx
GerberI: Done

src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -65,7 +65,7 @@ export const MainFilterList: DbKey[] = [
     "federalState",
 ];
 
-export const subFiltersList: string[] = [
+export const microorganismSubFiltersList: string[] = [
```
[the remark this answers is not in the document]
MicheleK17: The underscore keys are only needed in the MatrixDetails so that the keys are unique. 
For example, the microorganism (parent) Salmonella (trigger) has the subfilter Serovar - this combination is unique and occurs only in this way. 
But ALL subfilters of Matrix (parent) are called Matrixdetail. So that they cannot be uniquely assigned to their triggers. That's why the trigger is written behind "martixdetail" to not lose this information.

src/locales/de/QueryPage.json
```diff
@@ -843,48 +843,221 @@
...
-        "Sackware und lose Ware": "Loose or packaged in sacks",
-        "frisch": "Frisch",
-        "insb- Feldsalat, Rucola, Spinat, Eisberg": "Insb. Feldsalat, Rucola, Spinat, Eisberg",
-        "vorgeschnitten, verpackt": "Pre-cut, packaged",
-        "Einzeltierprobe": "Sngle animal sample",
-        "Poolprobe": "Poolprobe",
-        "wärmebehandelt": "Wärmebehandelt",
-        "lose oder verpackt; geschnitten oder Topfware": "Loose or packaged; pre-cut or in pots",
-        "gekühlt": "Chilled",
+    "matrixDetail__(Hals)haut": {
```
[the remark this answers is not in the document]
MicheleK17: I will discuss it with Dominic.

src/app/Main/Body/QueryPage/QueryPageContent/Parameter/ParameterList-Layout.component.tsx
```diff
@@ -9,10 +9,24 @@ const listStyle = css`
 const listHeaderStyle = css`
     position: relative;
 `;
+const spaceStyle = css`
+    padding: 0;
+    margin: 0;
+`;
+const parameterValue = css`
```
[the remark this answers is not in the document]
MicheleK17: Done.

src/app/Shared/Model/Client_Isolate.model.ts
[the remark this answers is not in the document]
GerberI: Thx, that makes sense. So "matrixDetail__*" can be read as "matrixDetail for *".

src/app/Main/Body/QueryPage/QueryPage-Container.component.tsx
```diff
@@ -175,6 +175,19 @@ export function QueryPageContainerComponent(): JSX.Element {
                 [keyName]: chooseSelectedFiltersService(selectedOption),
             },
         };
+        subFilters.forEach((subFilter) => {
```
GerberI, line 178: Some general hint why this peace of code and your data model is not the best one:
- You use specific information about the data here: "microorganism" and "matrix". This information is also used in different other parts of your code, so adding a new sub filters would become a difficult task.
- You mix the generic data subFilterTrigger with the specific data. Actually subFilterTriggers are bound to specific filters but this dependency is not visible in this peace of code. So the code looks like it is general but it is not. A more clean way would be to check the subfilters specific to a filter separately. But this would need also some refactoring of the model.

GerberI (approved):
Commit: Remove subFilter-values if parentFilter is deleted
As managing filters becomes more complex in each commit, it would be useful to separate all filter logic to a FilterManagingService. This service could provide methods to add/remove/change filters. The methods would get your whole filter structure as input and return a new modified filter structure as output. This is to encapsulate all filter specific code. You can write a refactoring ticket for this one.

src/app/Main/Body/QueryPage/Services/ContainerServices/fetchInitialData.service.tsx
```diff
@@ -42,7 +42,9 @@ export async function fetchInitialDataService(): Promise<{
         const adaptedSubFilters: ClientSingleFilterConfig[] = [];
         adaptedFilterProp.filters.forEach((adaptedFilter) => {
             if (adaptedFilter.parent !== undefined) {
-                adaptedSubFilters.push(adaptedFilter);
+                if (adaptedFilter.values.length > 1) {
```
GerberI, line 45: What is the meaning of sub-filters with only one value returned by the API?
In general it is good to check the data from the API before feeding it into your app.

GerberI (approved):
Commit: Only display subFilter with multiple values

GerberI (approved):
Commit: Rename style

GerberI:
Can be merged

src/app/Shared/Model/Client_Isolate.model.ts
```diff
@@ -119,6 +119,9 @@ export const matrixSubFiltersList: string[] = [
 export const allSubFiltersList =
     microorganismSubFiltersList.concat(matrixSubFiltersList);
 
+export const allFiltersList: (string | DbKey)[] =
+    allSubFiltersList.concat(MainFilterList);
```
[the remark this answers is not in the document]
MicheleK17: Done

PR #400, opened 2021-12-01
Define URLParams new for each parameter

MicheleK17:
Ticket: #399


dptolle (approved):

GerberI (approved):
Commit: Define URLParams new for each parameter
Sorry, seems i missed that one in a previous review.
As i stated some time ago, it would be better to assemble the required params from the url instead of deleting the not required ones. That would have made this bug impossible.

PR #603, opened 2023-09-12
Adding artifact to workflow

mahtabiltarabian:
#571 Adding artifact to workflow


mahtabiltarabian:
adding artifact to workflow(storing)

PR #683, opened 2023-11-10
662 adjust styling of the filter side panel

ckchinmay:
adjust styling of the filter side panel

### DIFF
--- a/src/locales/de/ExplanationPage.json
+++ b/src/locales/de/ExplanationPage.json
@@ -4,6 +4,7 @@
   "Data_Download": "Daten",
   "Search": "Suchen",
   "Filter": "Filtern",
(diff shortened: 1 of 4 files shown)


src/locales/de/ExplanationPage.json
```diff
@@ -4,6 +4,7 @@
   "Data_Download": "Daten",
   "Search": "Suchen",
   "Filter": "Filtern",
+  "Filter_Settings": "Einstellungen Filtern",
```
dptolle, line 7: Should be "Filter Einstellungen"

dptolle (changes requested):
There are merge conflicts that need to be taken care off

PR #730, opened 2024-01-12
add prevalence page same evaluation

mahtabiltarabian:


dptolle (commented):
You can probably create generalized components.
Let's discuss this on Monday

PR #756, opened 2024-02-15
Added Touch event handler for tooltip

mahtabiltarabian:
Added touch event handlers and  Added onTouchStart to handle touch events


dptolle (commented):
Why is this needed?

PR #801, opened 2024-03-27
implement icon for copy title to clipboard

mahtabiltarabian:
If clicking on the titles triggers an action (like opening a box with details), then the text is likely not selectable because it's functioning as a button or link, not as static text. In such cases, the text is often part of an interactive element designed for user interaction rather than text selection.,,, for this i implement icon for copy title to clipboard
image


dptolle (approved):

tasjacrease:
Can we move the icon to the end of the title as discussed in the meeting?

tasjacrease:
On the dev-server the copy icon is overlapping with the text of the title.

PR #1019, opened 2024-07-16
Error message link is on the Text zoonotify support  only

mahtabiltarabian:


src/app/prevalence/components/PrevalenceMainContent.tsx
```diff
@@ -27,6 +27,11 @@ const PrevalenceMainContent: React.FC<PrevalenceMainContentProps> = ({
               )}`
             : "";
 
+    // Extracting the message and explicitly making "zoonotify support" a link
+    const errorMessage = t("error notAllDataRetrieved");
```
dptolle, line 31: Please don't do string manipulation like this.  Adjust the entries in the localization json )possibly splitting them or parameterizing them)

dptolle (changes requested):

PR #1635, opened 2026-02-10
Id

mahtabiltarabian:
###  **Fix**

  I use entity names in URLs instead of documentId. Names are stable across re-imports. Internal state and API queries continue to use documentId — only the URL serialization layer changes.

  ### **Changes**

  **1. PrevalenceDataContext.tsx**
  - Added two helper functions: resolveUrlValueToDocId (URL name → docId when reading) and resolveDocIdToName (docId →   
  name when writing)
  - Added an optionsRef to hold current filter options synchronously for URL resolution
  - Writing URL: Before appending filter values to the URL, each docId is converted to its name via resolveDocIdToName   
  - Reading URL: On page load, URL values are resolved back to docIds via resolveUrlValueToDocId
  - Language change: fetchOptions() is always called first so the ref has current-locale names before the URL is
  rewritten

  **2. SubstanceDetail.tsx**
  - Added buildDocIdToNameMap / buildNameToDocIdMap helpers that build lookup maps from the resistance data
  - encodeStateToParam: Before compressing the payload, filter docIds are converted to names
  - readStateFromUrlCompressed: After decompressing, names are resolved back to docIds
  - getComboIdKey: Changed from documentId to name for stable combination keys
  - All 7 call sites updated to pass resistanceRawData for the conversion

  **3. TrendDetails.tsx** 
  - Added the same buildDocIdToNameMap / buildNameToDocIdMap / resolveUrlValueToDocIdTrend helpers
  - updateTrendFilterUrl: Converts docIds to names when writing URL params
  - readTrendFilterStateFromUrl: Resolves names back to docIds when reading URL params
  - Added hydratedFromUrlRef so the URL is only read once on initial load. On language change, the existing filter state 
  (docIds) is preserved and only the URL is rewritten with new-locale names
  - After language refetch, filteredFullData is recomputed so cascading dropdown options update immediately

### DIFF
--- a/src/app/antibiotic_resistance/pages/TrendDetails.tsx
+++ b/src/app/antibiotic_resistance/pages/TrendDetails.tsx
@@ -40,147 +40,16 @@ import { SidebarComponent } from "../../shared/components/layout/SidebarComponen
 
 import { TrendChart } from "./TrendChart";
 import type { SelectChangeEvent } from "@mui/material/Select";
-
-// --- All filter option keys
-type FilterKey =
-    | "samplingYear"
-    | "antimicrobialSubstance"
-    | "specie"
-    | "superCategorySampleOrigin"
-    | "sampleOrigin"
-    | "samplingStage"
-    | "matrixGroup"
-    | "matrix";
-
-type FilterOption = {
-    id: string;
-    name: string;
-    documentId: string; // shared cross-language!
-};
-
-const emptyFilterState: Record<FilterKey, string[]> = {
-    samplingYear: [],
-    antimicrobialSubstance: [],
-    specie: [],
-    superCategorySampleOrigin: [],
-    sampleOrigin: [],
-    samplingStage: [],
-    matrixGroup: [],
-    matrix: [],
-};
-
-export interface ResistanceApiItem {
-    id: number;
-    samplingYear: number;
-    superCategorySampleOrigin?: {
-        id: number;
-        name: string;
-        documentId: string;
-    } | null;
-    sampleOrigin?: { id: number; name: string; documentId: string } | null;
-    samplingStage?: { id: number; name: string; documentId: string } | null;
-    matrixGroup?: { id: number; name: string; documentId: string } | null;
-    matrix?: { id: number; name: string; documentId: string } | null;
-    antimicrobialSubstance?: {
-        id: number;
-        name: string;
-        documentId: string;
-    } | null;
-    specie?: { id: number; name: string; documentId: string } | null;
-    resistenzrate: number;
-    anzahlGetesteterIsolate: number;
-    anzahlResistenterIsolate: number;
-    minKonfidenzintervall: number;
-    maxKonfidenzintervall: number;
-}
-
-/** Get the relation object for a given filter key from a data item */
-function getRelObjectTrend(
-    item: ResistanceApiItem,
-    key: FilterKey
-): { name: string; documentId: string } | null {
-    switch (key) {
-        case "specie":
-            return item.specie ?? null;
-        case "superCategorySampleOrigin":
-            return item.superCategorySampleOrigin ?? null;
-        case "sampleOrigin":
-            return item.sampleOrigin ?? null;
-        case "samplingStage":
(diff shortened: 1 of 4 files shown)


src/app/antibiotic_resistance/pages/TrendDetails.tsx
```diff
@@ -69,73 +69,222 @@ const emptyFilterState: Record<FilterKey, string[]> = {
...
+    specie?: { id: number; name: string; documentId: string } | null;
+    resistenzrate: number;
+    anzahlGetesteterIsolate: number;
+    anzahlResistenterIsolate: number;
+    minKonfidenzintervall: number;
+    maxKonfidenzintervall: number;
+}
+
+/** Get the relation object for a given filter key from a data item */
+function getRelObjectTrend(
```
dptolle: This code is duplicated.  Consider externalising it into a Helper function or find a more Object oriented solution

src/app/antibiotic_resistance/pages/TrendDetails.tsx
```diff
@@ -69,73 +69,222 @@ const emptyFilterState: Record<FilterKey, string[]> = {
...
+            return item.matrix ?? null;
+        case "antimicrobialSubstance":
+            return item.antimicrobialSubstance ?? null;
+        default:
+            return null;
+    }
+}
+
+/** Build docId->name and name->docId maps for all filter keys */
+function buildDocIdToNameMapTrend(
```
dptolle: This code is duplicated. Consider externalising it into a Helper function or find a more Object oriented solution

src/app/antibiotic_resistance/pages/TrendDetails.tsx
```diff
@@ -69,73 +69,222 @@ const emptyFilterState: Record<FilterKey, string[]> = {
...
+            const obj = getRelObjectTrend(item, k);
+            if (obj?.documentId && obj?.name) m.set(obj.documentId, obj.name);
+        }
+        result[k] = m;
+    }
+    result.samplingYear = new Map<string, string>();
+    return result;
+}
+
+function buildNameToDocIdMapTrend(
```
dptolle: This code is duplicated. Consider externalising it into a Helper function or find a more Object oriented solution

src/app/antibiotic_resistance/pages/TrendDetails.tsx
```diff
@@ -69,73 +69,222 @@ const emptyFilterState: Record<FilterKey, string[]> = {
...
+            if (obj?.name && obj?.documentId) m.set(obj.name, obj.documentId);
+        }
+        result[k] = m;
+    }
+    result.samplingYear = new Map<string, string>();
+    return result;
+}
+
+/** Resolve a URL value (name or old docId) to docId with backwards compat */
+function resolveUrlValueToDocIdTrend(
```
dptolle: This code is duplicated. Consider externalising it into a Helper function or find a more Object oriented solution

src/app/antibiotic_resistance/pages/TrendDetails.tsx
```diff
@@ -69,73 +69,222 @@ const emptyFilterState: Record<FilterKey, string[]> = {
     matrix: [],
 };
 
+export interface ResistanceApiItem {
```
dptolle: This interface shouldn't really be in TrendsDetails.tsx if it is also used in SubstanceDetail.tsx.  Please Externalise.

src/app/prevalence/components/PrevalenceDataContext.tsx
```diff
@@ -114,6 +115,24 @@ const writeChartMicroToUrl = (microName: string): void => {
     }
 };
 
+/** URL <-> docId resolution helpers */
+function resolveUrlValueToDocId(
```
dptolle, line 120: This code is not just duplicated, but also underwent a slight, and probably unnecessary, change in function signature. Consider using the (already externalised, as mentioned in previous comments) Helpderfunction?Object

dptolle (changes requested):

dptolle (approved):

